PR #1, opened 2021-12-06
create owner file

bmeng:


georgettica:
shouldn't the owner file be present in the github.com/openshift/release ?

PR #5, opened 2022-01-13
onboarding ocm-agent into codecov.io

a7vicky:
### What type of PR is this?
onboarding com-agent into codecov.io


mrbarge:
I was unable to get `make coverage` working as-is, this seems to be due to the `codecov.sh` script using a `go-test` target which the Makefile currently does not have.

What I'd recommend is the following changes:
- remove the existing `coverage` target (https://github.com/openshift/ocm-agent/blob/4b83fa8ce0dd03e783cd3975080f97b287a9b11e/Makefile#L101) and the script that it calls (`hack/test.sh`)
- rename the existing `test` target (https://github.com/openshift/ocm-agent/blob/4b83fa8ce0dd03e783cd3975080f97b287a9b11e/Makefile#L45) to `go-test` 
- introduce a new `test` target that just has `go-test` as a dependency (this aligns with boilerplate):

```
.PHONY: test
test: go-test
```

After doing that, `make coverage` should succeed and result in a `coverage.out`.

a7vicky:
Thank you @mrbarge  for the review. I have updated the Makefile and `make coverage` is functional now.

mrbarge:
Perfect, thanks!
/lgtm

PR #7, opened 2022-01-18
Correct skopeo argument in build-push target

mrbarge:
### What type of PR is this?
Bug

The `build-push` Make target was using an invalid transport means of `container-storage`, it should be `containers-storage`.


ravitri:
/lgtm

Liked the reference https://www.redhat.com/sysadmin/7-transports-features regarding it

PR #9, opened 2022-02-02
[OSD-9797] Add AlertManager receiver webhook endpoint

T0MASD:
### What type of PR is this?

Add AlertManager receiver webhook endpoint

_Fixes_ [OSD-9797](https://issues.redhat.com/browse/OSD-9797)

- [x ] Tested locally


T0MASD:
As per conversation on slack I'm looking into updating the tests to use ginkgo framework

cmd/ocm-agent/main.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/openshift/ocm-agent/hack"
```
georgettica, line 9: Unrelated, change hack to fake?

bmeng:
if this is AM specific, should we add the `AM` name either in the package name or the go file name?

T0MASD:
> if this is AM specific, should we add the `AM` name either in the package name or the go file name?
@bmeng thanks for taking a look at this PR. It's not strictly Alert Manager specific, just happens that Alert Manager receiver handler is implemented first. Any future handlers for webhook receivers are expected in handlers.go of webhookreceiver package

cmd/ocm-agent/main.go
T0MASD: Thanks for your message Ron. I believe this will be removed completely as part of [OSD-8943](https://issues.redhat.com/browse/OSD-8943) is implemented. Please follow up or resolve this discussion

pkg/webhookreceiver/handler.go
```diff
@@ -0,0 +1,32 @@
...
+	h.Route(r.NewRoute().HandlerFunc(h.Func))
+}
+
+// Alert Manager receiver response
+type AMReceiverResponse struct {
+	Status string
+}
+
+// Set webhook receiver path
+const AMReceiverPath = "/alertmanager-receiver"
```
ravitri, line 29:
```suggestion
const AMReceiverPath = "/receiver"
```
Should we simplify the URL path without hyphens? Reference [link](https://swagger.io/resources/articles/best-practices-in-api-design/#nouns-describe-urls-better-3)

ravitri: If there are going to be other receivers in future too then makes sense to have "alertmanager" in path.. But have rarely seen any API URL path with "hyphen" in it.

pkg/webhookreceiver/handlers_test.go
```diff
@@ -0,0 +1,82 @@
+package webhookreceiver
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"testing"
+
+	"github.com/gorilla/mux"
+	. "github.com/onsi/ginkgo"
```
mrbarge, line 10: We've just made the switch to Ginkgov2 in OSDE2E in the last week.. since this would be the introduction of Ginkgo into the OA codebase, might be good to start off with v2 from the get-go.  

On the whole this should be pretty easy and not need many (hopefully no) changes.

https://onsi.github.io/ginkgo/MIGRATING_TO_V2 for the details.

pkg/webhookreceiver/handlers.go
```diff
@@ -0,0 +1,38 @@
...
+			if r != nil && r.Method != http.MethodPost {
+				http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
+				return
+			}
+			// process request
+			var alertData AMReceiverData
+			json.NewDecoder(r.Body).Decode(&alertData)
+			go processAMReceiver(alertData)
+
+			// write response
```
mrbarge, line 27: Even though it would happen by default, I think it is good for code readability to explicitly set the return HTTP status code here with a `w.WriteHeader` call to set it to `http.StatusOK` to signify that we consider this reply an indication of success.

T0MASD: Added w.WriteHeader(http.StatusOK)

pkg/webhookreceiver/handlers_test.go
T0MASD: Added ginkgo v2

pkg/webhookreceiver/handlers.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+		Func: func(w http.ResponseWriter, r *http.Request) {
+			// validate request
+			if r != nil && r.Method != http.MethodPost {
+				http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
+				return
+			}
+			// process request
+			var alertData AMReceiverData
+			json.NewDecoder(r.Body).Decode(&alertData)
```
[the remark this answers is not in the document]
T0MASD: added additional error handling

pkg/webhookreceiver/handler.go
T0MASD: The reason for choosing this path was because this is what was used for POC, additionally there's some anecdotal evidence that hyphens is preferred way to separate the verbs in rest api path. I personally like the name as it identifies path as alertmanager

ravitri: Agreed.. We can revisit this in future when required too.. For now, resolving the comment. Thanks Tomas!

mrbarge:
Feel free to unhold when you are satisfied that this is good to go. 
/hold
/lgtm

mrbarge:
I successfully tested this with a on-cluster AlertManager connection to an `ocm-agent` deployed by the latest `ocm-agent-operator`.

```
time="2022-02-13T10:57:02Z" level=info msg="Process alert data" AMReceiverData="{ocm-agent firing [{firing map[alertname:ConfigureAlertmanagerOperatorOfflineSRE namespace:openshift-monitoring prometheus:openshift-monitoring/k8s service:configure-alertmanager-operator severity:critical] map[] 2022-02-13 10:27:02.577 +0000 UTC 0001-01-01 00:00:00 +0000 UTC https://prometheus-k8s-openshift-monitoring.apps.mbarg-49.dr1m.s1.devshift.org/graph?g0.expr=absent%28up%7Bservice%3D%22configure-alertmanager-operator%22%7D%29&g0.tab=1 3be6ab161d896f54}] map[] map[alertname:ConfigureAlertmanagerOperatorOfflineSRE namespace:openshift-monitoring prometheus:openshift-monitoring/k8s service:configure-alertmanager-operator severity:critical] map[] https://alertmanager-main-openshift-monitoring.apps.mbarg-49.dr1m.s1.devshift.org}"
```

Works well, thanks for the addition.


/lgtm

Makefile
```diff
@@ -51,7 +51,7 @@ clean:
 
 .PHONY: serve
 serve:
-	$(AT)go run ./cmd/ocm-agent/main.go
+	$(AT)go run ./cmd/ocm-agent/main.go serve --access-token "$TOKEN" --services "$SERVICE" --ocm-url "https://sample.example.com"
```
ravitri, line 54: Regarding this, I wonder if we should add this in Makefile after the CLI has env variable or defaults support. For now, is this change required as part of this PR?

pkg/cli/serve/serve.go
```diff
@@ -54,13 +53,13 @@ var (
 
 	serviceExample = templates.Examples(`
 	# Start the OCM agent server
-	ocm-agent server --access-token "$TOKEN" --services "$SERVICE" --ocm-url "https://sample.example.com"
+	ocm-agent serve --access-token "$TOKEN" --services "$SERVICE" --ocm-url "https://sample.example.com"
```
ravitri, line 56: Nice catch :) Thanks! I missed correcting this in previous PR

PR #10, opened 2022-02-07
OSD-9709 - Adding ocm-agent serve CLI implementation

ravitri:
### What type of PR is this?
_(feature)_

To create a CLI implementation to be able to start ocm-agent server on accepting the access-token, services and OCM URL.

_Fixes #_ [OSD-9709](https://issues.redhat.com/browse/OSD-9709)


pkg/cli/serve/serve.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *ServeOptions) Run() error {
+
+	port := 8081
```
mrbarge, line 91: I'm thinking it'd be good if the value of this was defined as a `const`, maybe in a dedicated `consts` package.
Thinking maybe it might be good to have this be a command-line option to `serve` too, come to think of it (but as an optional argument that defaults to the constant).

But we can wait until #9 and #10 are both merged (and whatever conflicts are ironed out) before worrying about that, it doesn't need to be sorted out in this PR.

ravitri: Thanks a lot for the feedback Matt!! Sure, will handle it in a separate PR soon.

ravitri: For now, moving `port` as const but will work on the const package and configurable port as a separate PR

mrbarge:
This all looks great to me.

a7vicky:
Tested and it's working fine.
/lgtm

pkg/cli/root.go
```diff
@@ -0,0 +1,47 @@
+package cli
```
bmeng, line 1: the file name here is a bit horrible? should better use the real cmd name `ocm-agent.go` IMO.

pkg/cli/serve/serve.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (l *level) Set(value string) error {
+	lvl, err := logrus.ParseLevel(strings.TrimSpace(value))
+	if err == nil {
+		*l = level(lvl)
+	}
+	return err
+}
+
+func (l *level) Type() string {
```
bmeng, line 33: what does this function do? seems it is not being used.

pkg/cli/serve/config.go
```diff
@@ -0,0 +1,40 @@
...
+
+// ReadFlagsFromFile checks for '@' prefix, if found try to read value from file.
+func ReadFlagsFromFile(cmd *cobra.Command, names ...string) error {
+	for _, name := range names {
+		value, err := cmd.Flags().GetString(name)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unable to fetch value for key %s\n", name)
+			os.Exit(1)
+		}
+		if strings.HasPrefix(value, "@") {
```
bmeng, line 20: why we have to have a `@` in each of the values?

mrbarge: This was something I requested as part of the OSD-9709 acceptance criteria.

If the argument is preceded by a "@" the agent infers the value supplied via the CLI is a filepath which contains the config value, and reads the value from there. 

If the argumnt is not, it infers the value supplied on the CLI. 

The former behaviour allows for the ability to mount a ConfigMap through to the container and read files in the way that configmap keys/values get mounted. The latter behaviour is easier for local testing.

Cluster Service works in a very similar way.

pkg/cli/serve/serve.go
ravitri: Hey Meng Bo! Nice catch.. I do remember this implementation was required somewhere while accepting CLI flags but maybe I was mistaken while testing something else.. Wonder why IDE didn't catch this.. Yup, removed this bit.. Checked the CLI again and seems to work well

pkg/cli/root.go
ravitri: Well, I tried following convention as per https://cobra.dev/#create-rootcmd. `main()` would execute the CLI's command and the root command would have the mapping of all the subcommands. I did come across different conventions from different repos though. Another example is https://github.com/openshift/osd-network-verifier/blob/main/cmd/root.go 

Let me know if this is ok. Else, we can change it too.

PR #11, opened 2022-02-17
OSD-8946 implement the metrics

bmeng:
### What type of PR is this?
_feature_

_Fixes #[OSD-8946](https://issues.redhat.com/browse/OSD-8946)_


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	OCMAgentTotalRequests = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "ocm_agent_total_requests",
```
dofinn, line 13: https://prometheus.io/docs/practices/naming/ we should follow prometheus conventions?

```suggestion
			Name: "ocm_agent_requests_total",
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+			Help: "A count of total requests to ocm agent service",
+		}, []string{})
+
+	OCMAgentByServiceRequests = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "ocm_agent_service_requests",
```
dofinn, line 19: given that I know the ocm-agent architecture, this metric is not entirely clear. 

Is it:

* a request to ocm-agent ?
* a request api.openshift.com?
* both?

Maybe we could hint to the operator reviewing these metrics which request this is?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+		prometheus.CounterOpts{
+			Name: "ocm_agent_failed_requests",
+			Help: "A count of total failed request to ocm agent service",
+		}, []string{})
+
+	OCMAgentRequestFailure = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "ocm_agent_request_failure",
+			Help: "Indicate that the request send to ocm agent failed",
+		}, []string{"path"})
```
dofinn, line 33: Could you please explain the difference between these?

dofinn (commented):
There is an option to add tests with promutil. An example can be found here: https://github.com/openshift/managed-upgrade-operator/blob/master/pkg/collector/collector_test.go

pkg/cli/serve/serve.go
```diff
@@ -64,11 +67,12 @@ var (
 )
 
 const (
-	port                = 8081
-	accessTokenFlagName = "access-token"
-	servicesFlagName    = "services"
-	ocmURLFlagName      = "ocm-url"
-	debugFlagName       = "debug"
+	servicePort         int    = 8081
+	metricsPort         int    = 8888
```
mrbarge, line 71: Although ultimately we can make the service port be different to that in the container, the operator sdk convention for the metrics port is `8383` (well, until 1.0, then it changed to 8080) so it might be worth using that for consistency.

pkg/metrics/metrics.go
bmeng: The `OCMAgentRequestFailure` is a gauge value, to indicate that the request to the ocm-agent is failed, and we are going to use this to alert SRE to fix the issue. The value will be change back to 0 once the issue is gone.
OCMAgentTotalFailedRequests is a counter, like statistics. To calculate the total failed request, can be also exported to external tools.

pkg/metrics/metrics.go
bmeng: This is the counter for the request to ocm-agent, but for different paths/endpoints.
Eg, we are supporting alertmanager as an endpoint for ocm-agent for now.
The counter here will count all the requests to the alertmanager endpoints. 
And if we are going to support the other endpoints in future. There will be a different counter for them. 
It just be used for statistics.

georgettica:
I love this as I need to implement an http server and now I have a tested code snippet.

Good job and good luck!

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+	OCMAgentByServiceRequests = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "ocm_agent_service_requests",
+			Help: "A count of total requests to ocm agent based on sub service",
+		}, []string{"path"})
+
+	OCMAgentTotalFailedRequests = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "ocm_agent_failed_requests",
+			Help: "A count of total failed request to ocm agent service",
```
mrbarge, line 26: Minor grammar tweak
```suggestion
			Help: "A count of total failed requests received by the OCM Agent service",
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+		}, []string{})
+
+	OCMAgentRequestFailure = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "ocm_agent_request_failure",
+			Help: "Indicate that the request send to ocm agent failed",
```
mrbarge, line 32: Minor grammar tweak
```suggestion
			Help: "Indicates that OCM Agent could not successfully process a request",
```

pkg/metrics/metrics.go
mrbarge: +1

mrbarge:
Gave this a test and it's all functionally working great. Left some comments of things I'd like to see tweaked (also +1 to Dom's comments), but other than that this looks nice.

mrbarge:
Thanks for the changes - seems like this addresses all the feedback; I gave it another test and it worked fine.
/lgtm

PR #14, opened 2022-03-07
OSD-10226 Add cluster ID and support viper

mrbarge:
### What type of PR is this?
Feature

OCM Agent needs to know the cluster ID; it's currently getting that injected into the ConfigMap used to supply config to the pod; now we need to make the agent aware that it has to read that config.

Additionally, `viper` has been introduced to enable easier retrieval of app configuration items, this will make subsequent tickets a lot easier to deal with when they need to access the application's config. eg.

```
viper.Get(config.SomeConfigKey)
```

[OSD-10226](https://issues.redhat.com/browse/OSD-10226)

### DIFF
--- a/pkg/cli/serve/serve.go
+++ b/pkg/cli/serve/serve.go
@@ -96,7 +96,7 @@ func NewServeCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&o.ocmURL, config.OcmURL, "", "", "OCM URL")
 	cmd.Flags().StringVarP(&o.services, config.Services, "", "", "OCM service name")
 	cmd.Flags().StringVarP(&o.accessToken, config.AccessToken, "t", "", "Access token for OCM")
-	cmd.Flags().StringVarP(&o.clusterID, config.ClusterID, "i", "", "Cluster ID")
+	cmd.Flags().StringVarP(&o.clusterID, config.ClusterID, "c", "", "Cluster ID")
 	cmd.PersistentFlags().BoolVarP(&o.debug, config.Debug, "d", false, "Debug mode enable")
 	viper.BindPFlags(cmd.Flags())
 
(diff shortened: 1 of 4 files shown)


pkg/cli/serve/serve.go
```diff
@@ -94,14 +93,17 @@ func NewServeCmd() *cobra.Command {
...
 	}
 
-	cmd.Flags().StringVarP(&o.ocmURL, ocmURLFlagName, "", "", "OCM URL")
-	cmd.Flags().StringVarP(&o.services, servicesFlagName, "", "", "OCM service name")
-	cmd.Flags().StringVarP(&o.accessToken, accessTokenFlagName, "t", "", "Access token for OCM")
-	cmd.PersistentFlags().BoolVarP(&o.debug, debugFlagName, "d", false, "Debug mode enable")
+	cmd.Flags().StringVarP(&o.ocmURL, config.OcmURL, "", "", "OCM URL")
+	cmd.Flags().StringVarP(&o.services, config.Services, "", "", "OCM service name")
+	cmd.Flags().StringVarP(&o.accessToken, config.AccessToken, "t", "", "Access token for OCM")
+	cmd.Flags().StringVarP(&o.clusterID, config.ClusterID, "i", "", "Cluster ID")
```
ravitri, line 99:
```suggestion
	cmd.Flags().StringVarP(&o.clusterID, config.ClusterID, "c", "", "Cluster ID")
```
Minor point.. Am thinking if 'c' for cluster is more intuitive than 'i' for cluster ID?

ravitri:
Tested CLI and overall lgtm. Just has one minor point regarding the short flag for cluster ID.

pkg/cli/serve/serve.go
mrbarge: That makes sense and would be more consistent with `ocm` and `rosa` CLIs - thanks I have made that change now.

PR #15, opened 2022-03-09
OSD-10224 Refactor to introduce k8s dependency

mrbarge:
### What type of PR is this?
Refactor

`ocm-agent`'s current structure and design does not lend itself to the injection of dependencies to the handlers, nor to testing with fake or mocked versions of those dependencies.

This is something that will be needed once we actually implement the webhook receiver and need to inject a k8s client, an OCM client, and so on. 

The handlers have been refactored to allow such design to be possible.

- None of the actual handler code is changed or removed
- None of the test code is changed or removed (but revised slightly to take advantage of the refactor)

As a demonstration to verify the refactor design, the webhook receiver handler now allows injection of a kube client which is initalized outside of the handler. Thus, this change introduces the controller-runtime as a dependency. The controller-runtime version chosen was deliberate because there are dependency clashes with more recent versions when OCM SDK is brought into the picture.

As such, `ocm-agent` now requires that have set an environment which will allow it to detect a kubeconfig, ie. by setting the `KUBECONFIG` environment variable. On-cluster this should already be the case, but if testing locally, it will now exit if it can't initialize a kube client. 

This change is necessary to allow [OSD-10224](https://issues.redhat.com/browse/OSD-10224) to be written.


go.mod
```diff
@@ -4,15 +4,21 @@ go 1.16
...
 	github.com/prometheus/alertmanager v0.23.0
 	github.com/prometheus/client_golang v1.11.0
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/cobra v1.3.0
 	github.com/spf13/pflag v1.0.5
-	k8s.io/kubectl v0.23.3
 	github.com/spf13/viper v1.10.1
+	k8s.io/apimachinery v0.22.1
+	k8s.io/cli-runtime v0.22.1 // indirect
+	k8s.io/client-go v0.22.1 // indirect
```
ravitri, line 19: We might need to pin the indirect deps under `replace` directive. Running `go mod tidy` for the clone of the PR is changing state of `go.mod` file for now.

mrbarge: Ahh, thanks for catching that - no, that's just a goof-up on my part. Fixed now.

pkg/cli/serve/serve.go
```diff
@@ -131,32 +131,32 @@ func (o *serveOptions) Run() error {
...
-	r := mux.NewRouter()
-
-	// Add healthcheck routes
-	healthcheck.Livez().AddRoute(r)
-	healthcheck.Readyz().AddRoute(r)
 	// create new router for metrics
 	rMetrics := mux.NewRouter()
 	rMetrics.Path("/metrics").Handler(promhttp.Handler())
 
-	r.Use(metrics.PrometheusMiddleware)
```
bmeng, line 144: Remove this will not call the `PromethuesMiddleware` for the default router. And the metrics cannot be collected as well.

Is this intentional?

bmeng:
Gave this a review and test. Except the one comment above. looks good.

pkg/cli/serve/serve.go
mrbarge: Thanks a lot for spotting that, that was unintended.

I have pushed a commit to add it back. However, I must ask why the Prometheus middleware would need to be  added to the default router? Shouldn't it be added to the metrics router which is actually serving the metrics endpoint?

I pushed a commit adding it to the metrics router and verified that Prometheus was able to crawl metrics successfully afterwards.

bmeng: The middleware running on default router to proceed and generate the metrics based on the request to default router.
The metrics router just exposes them to metrics port.

mrbarge: Fair enough - I have moved it to the default router.

PR #18, opened 2022-03-14
OSD-10204 Refactor ocm_agent_requests_total metric

anjoshi24:
Signed-off-by: Aniket Joshi <anjoshi@redhat.com>

### What type of PR is this?
Refactor

This PR change refactors existing logic of ocm_agent_requests_total metric to exclude unnecessary path requests count for more clarity. 

[OSD-10204](https://issues.redhat.com/browse/OSD-10204)

_Fixes #_

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
+	"github.com/openshift/ocm-agent/pkg/handlers"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -73,9 +74,10 @@ func (rw *responseWriter) WriteHeader(code int) {
 func PrometheusMiddleware(ph http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 
-		metricRequestsTotal.WithLabelValues().Inc()
-
 		path, _ := mux.CurrentRoute(r).GetPathTemplate()
+		if path != handlers.LivezPath && path != handlers.ReadyzPath {
+			metricRequestsTotal.WithLabelValues().Inc()
+		}
 		metricRequestsByService.WithLabelValues(path).Inc()
```
bmeng, line 81: I think we should put this in the if condition as well?
Since collect the metric for the livez and readyz is also useless.

anjoshi24: Thanks for the review, I also had a similar thought that we might not want to record the path requests to livez & readyz. But initially it posed as a good testing condition for clarification. 
I will make this change.

anjoshi24: Made this change and tested it against a cluster.

PR #21, opened 2022-03-15
OSD-10339 Do not send a notification if within the window

mrbarge:
### What type of PR is this?
Feature

Proposed implementation of [OSD-10339](https://issues.redhat.com/browse/OSD-10339) to not send a notification if it has already been sent within the do-not-resend window. Needs a bunch of tests still to verify correctness.

[OSD-10339](https://issues.redhat.com/browse/OSD-10339)

### DIFF
--- a/pkg/handlers/webhookreceiver.go
+++ b/pkg/handlers/webhookreceiver.go
@@ -120,7 +120,7 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 	}
 
 	// Can the alert be mapped to an existing notification definition?
-	notification, managedNotifications, err := getNotification(alert.Labels[AMLabelManagedNotification], mnl)
+	notification, managedNotifications, err := getNotification(alert.Labels[AMLabelTemplateName], mnl)
 	if err != nil {
 		log.WithError(err).WithField(LogFieldAlert, alert).Warning("an alert fired which no notification template definition exists for")
 		return err
--- a/pkg/handlers/webhookreceiver.go
+++ b/pkg/handlers/webhookreceiver.go
@@ -24,6 +24,13 @@ const (
 	AMLabelAlertName           = "alertname"
 	AMLabelTemplateName        = "managed_notification_template"
 	AMLabelManagedNotification = "send_managed_notification"
+
+	LogFieldNotificationName    = "notification"
+	LogFieldResendInterval      = "resend_interval"
+	LogFieldAlertname           = "alertname"
+	LogFieldAlert               = "alert"
+	LogFieldIsFiring            = "is_firing"
+	LogFieldManagedNotification = "managed_notification_cr"
 )
 
 type WebhookReceiverHandler struct {
@@ -82,7 +89,7 @@ func (h *WebhookReceiverHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 func (h *WebhookReceiverHandler) processAMReceiver(d AMReceiverData, ctx context.Context) *AMReceiverResponse {
 	log.WithField("AMReceiverData", d).Info("Process alert data")
 
-	// Let's get all ManagedNotifications
+	// Let's get all ManagedNotifications in the
 	mnl := &oav1alpha1.ManagedNotificationList{}
 	listOptions := []client.ListOption{
 		client.InNamespace("openshift-ocm-agent-operator"),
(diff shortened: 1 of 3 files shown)


pkg/handlers/webhookreceiver.go
```diff
@@ -93,49 +100,91 @@ func (h *WebhookReceiverHandler) processAMReceiver(d AMReceiverData, ctx context
...
-			log.WithField("alertname", alertname).Debug("ignoring alert with no send_managed_notification label")
-			continue
-		}
+// processAlert handles the pre-check verification and sending of a notification for a particular alert
+// and returns an error if that process completed successfully or false otherwise
+func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alpha1.ManagedNotificationList) error {
+	// Should this alert be handled?
+	if !isValidAlert(alert) {
+		log.WithField(LogFieldAlert, alert).Info("alert does not meet valid criteria")
+		return fmt.Errorf("alert does not meet valid criteria")
```
bmeng, line 119: we log the same message twice?

bmeng: another question, should we return error here? or we just log this as Info and return nil?

pkg/handlers/webhookreceiver.go
```diff
@@ -93,49 +100,91 @@ func (h *WebhookReceiverHandler) processAMReceiver(d AMReceiverData, ctx context
...
+	}
 
-		notification, mn, err := getNotification(notificationName, mnl)
-		if err != nil {
-			log.WithError(err).WithField("alertname", alertname).Warning("an alert fired which no notification template definition exists for")
-			continue
-		}
-		log.WithFields(log.Fields{"notification": notification.Name, "alertname": alertname}).Info("Found a notification")
+	// Has a servicelog already been sent and we are within the notification's "do-not-resend" window?
+	canBeSent, err := managedNotifications.CanBeSent(notification.Name)
```
bmeng, line 130: We should have a separate card to implement/improve the CanBeSent func?
From what I am reading now, it is doing simple thing only. Not sure if it could cover our use case

pkg/handlers/webhookreceiver.go
T0MASD: should this return
````
		return &AMReceiverResponse{Error: err, Status: "alert does not meet valid criteria", Code: http.StatusInternalServerError}
	}
````

File: pkg/handlers/webhookreceiver.go (hunk 2 above)
Comment (mrbarge) [reply pkg/handlers/webhookreceiver.go:130]: Can you please clarify what you think the current `CanBeSent` implementation is missing? I consider it a complete implementation as-is, but have I forgotten something?

File: pkg/handlers/webhookreceiver.go (hunk 1 above)
Comment (T0MASD) [reply pkg/handlers/webhookreceiver.go:119]: should this return
````
		return &AMReceiverResponse{Error: err, Status: "alert does not meet valid criteria", Code: http.StatusInternalServerError}
	}
````

Comment (mrbarge) [reply pkg/handlers/webhookreceiver.go:119]: We could, my main reasoning was whether we ultimately want to do any custom logic back at the `processAlert()` level if the alert is invalid rather than just not needing a notification sent. I originally just had it returning a `bool` to indicate whether a service log was sent or not, then changed it to an error at the last minute. 

Error handling from bad AM input in general is something we still need to address and think about in OA.

Comment (mrbarge) [reply pkg/handlers/webhookreceiver.go:119]: @T0MASD I disagree, this part of the code should not be caring about the structure of data that is sent back to AlertManager. That should be the handler's job. These functions should be able to run and tested without them needing to know they're interacting with an external client.

File: pkg/handlers/webhookreceiver.go [hunk 3]
```diff
@@ -93,49 +100,91 @@ func (h *WebhookReceiverHandler) processAMReceiver(d AMReceiverData, ctx context
...
+	}
 
-		// Ignore alerts which don't have a notification defined
-		notificationName, foundNotification := alert.Labels[AMLabelTemplateName]
-		if !foundNotification {
-			log.WithField("alertname", alertname).Error("alert does not have template defined")
-			continue
-		}
+	// Can the alert be mapped to an existing notification definition?
+	notification, managedNotifications, err := getNotification(alert.Labels[AMLabelManagedNotification], mnl)
```
ravitri, line 123:
```suggestion
	notification, managedNotifications, err := getNotification(alert.Labels[AMLabelTemplateName], mnl)
```
Testing this, came across below error and realised that the label is incorrect:
```
time="2022-03-15T11:55:31Z" level=error msg="an alert could not be successfully processed" error="matching managed notification not found for true"
```

After fixing the label, got below log:
```
time="2022-03-15T13:20:08Z" level=info msg="will send servicelog for notification" notification=TestTemplate
```

PR #22, opened 2022-03-16
[OSD-10340] update the status for managed notification object

bmeng:
### What type of PR is this?
_feature_

Update the managednotification status based on the status transition

_Fixes #OSD-10340_

### DIFF
--- a/pkg/handlers/webhookreceiver.go
+++ b/pkg/handlers/webhookreceiver.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	sdk "github.com/openshift-online/ocm-sdk-go"
@@ -15,6 +16,8 @@ import (
 	oav1alpha1 "github.com/openshift/ocm-agent-operator/pkg/apis/ocmagent/v1alpha1"
 	"github.com/openshift/ocm-agent/pkg/config"
 	"github.com/openshift/ocm-agent/pkg/ocm"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -149,7 +152,7 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 	}
 
 	// Update the notification status to indicate a servicelog has been sent
-	err = h.updateNotificationStatus(notification, managedNotifications)
+	err = h.updateNotificationStatus(notification, managedNotifications, true)
 	if err != nil {
 		log.WithFields(log.Fields{LogFieldNotificationName: notification.Name, LogFieldManagedNotification: managedNotifications.Name}).WithError(err).Error("unable to update notification status")
 		return err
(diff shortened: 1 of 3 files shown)


bmeng:
/hold
need rebase after https://github.com/openshift/ocm-agent-operator/pull/31 merged

bmeng:
Code rebased and test on staging cluster.

Scenarios: 
1. When the alert first fired
```
  status:
    notifications:
    - conditions:
      - lastTransitionTime: "2022-03-16T09:19:44Z"
        reason: Alert is firing
        status: "True"
        type: AlertFiring
      - lastTransitionTime: "2022-03-16T09:19:44Z"
        reason: Alert resolved
        status: "False"
        type: AlertResolved
      - lastTransitionTime: "2022-03-16T09:19:44Z"
        reason: Service log sent
        status: "True"
        type: ServiceLogSent
      name: TestTemplate
      serviceLogSentCount: 1
```

2. When the resend window passed
```
  status:
    notifications:
    - conditions:
      - lastTransitionTime: "2022-03-16T09:19:44Z"
        reason: Alert is firing
        status: "True"
        type: AlertFiring
      - lastTransitionTime: "2022-03-16T09:19:44Z"
        reason: Alert resolved
        status: "False"
        type: AlertResolved
      - lastTransitionTime: "2022-03-16T09:34:45Z"
        reason: Service log sent
        status: "True"
        type: ServiceLogSent
      name: TestTemplate
      serviceLogSentCount: 2

```

The alert resolved can be handled by the status update, can be used directly when we implement the alert resolving in the `processAlert()`

pkg/handlers/webhookreceiver.go
```diff
@@ -250,18 +253,52 @@ func (h *WebhookReceiverHandler) updateNotificationStatus(n *oav1alpha1.Notifica
...
+					status.SetStatus(oav1alpha1.ConditionServiceLogSent, "Service log sent", corev1.ConditionTrue, timeNow)
+				} else {
+					// Status transition is Firing to Resolved
+					// Update the condition status and timestamp for AlertFiring
+					// Update the condition status and timestamp for AlertResolved
+					// Update the timestamp for the ServiceLogSent
+					status.SetStatus(oav1alpha1.ConditionAlertFiring, "Alert is firing", corev1.ConditionFalse, timeNow)
+					status.SetStatus(oav1alpha1.ConditionAlertResolved, "Alert resolved", corev1.ConditionTrue, timeNow)
+					status.SetStatus(oav1alpha1.ConditionServiceLogSent, "Service log sent", corev1.ConditionTrue, timeNow)
+				}
```
ravitri, line 290: So from the flow I see that `processAMReceiver` method would called `processAlert` method in a loop for only the firing alerts. This is why, the newly added bool value for `updateNotificationStatus` is set to `true` always. Considering that, will this highlighted code in the `else` block be ever reached?

mrbarge: My guess is this is being written in advance of OSD-10225 to implement `resolved` handling (which would make the same call, but with `false`), but Meng Bo will be in the best position to advise.

bmeng: Yes, the block for the `!firing` is used for the resolve part in OSD-10225.
I am just thinking consider them together will be easier from the logic over all.

And when we are going to implement the resolve alert logic.
We can just call the same function `updateNotificationStatus()` with a `false` in the end, just like Matt mentioned.

We may need to do some small tweak in `processAlert()` function then if we want to use it in the loop `for _, alert := range d.Alerts.Resolved()` like what we are doing now in L107 for firing alerts.

PR #29, opened 2022-03-29
[OSD-10315] setup metrics for alertmanager webhook receiver

bmeng:
### What type of PR is this?
_feature/refactor_

count the total service log sent based on the template
update the gauge values once we got the request to ocm agent or response from service logs failed
reset the gauge values once we got the succeeded request to ocm agent or response from service logs 

_Fixes #OSD-10315_


pkg/metrics/metrics.go
```diff
@@ -75,7 +81,7 @@ func PrometheusMiddleware(ph http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 
 		path, _ := mux.CurrentRoute(r).GetPathTemplate()
-		if path != handlers.LivezPath && path != handlers.ReadyzPath {
+		if path != "/livez" && path != "/readyz" {
```
mrbarge, line 84: Any particular reason to stop referring to the constants here?

pkg/handlers/webhookreceiver.go
```diff
@@ -104,8 +106,11 @@ func (h *WebhookReceiverHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	if err != nil {
 		log.Errorf("Failed to write to response: %s\n", err)
 		http.Error(w, "Failed to write to response", http.StatusInternalServerError)
+		metrics.SetRequestMetricFailure(WebhookReceiverPath)
 		return
 	}
+
+	metrics.ResetMetric("metricRequestFailure")
```
mrbarge, line 113: Once there's a string being used to in two separate functions which impacts logic flow, it might be a sign to turn it into a constant that both functions can refer to. Maybe turn this string and `"metricResponseFailure"` to constants which this func and the `ResetMetric` one can use?

mrbarge:
Left a couple of minor questions/comments, but I tested this on a cluster and it worked well.

T0MASD:
There are few lint errors as seen in `make lint`

```
pkg/cli/serve/serve.go:198:6: `fakeCallServiceLog` is unused (deadcode)
func fakeCallServiceLog(w http.ResponseWriter, r *http.Request) {
     ^
pkg/cli/serve/serve.go:204:6: `fakeReqFailure` is unused (deadcode)
func fakeReqFailure(w http.ResponseWriter, r *http.Request) {
     ^
pkg/metrics/metrics.go:59:22: Error return value of `prometheus.Register` is not checked (errcheck)
		prometheus.Register(m)
		                   ^
pkg/handlers/webhookreceiver.go:314:20: Error return value of `status.SetStatus` is not checked (errcheck)
			status.SetStatus(oav1alpha1.ConditionAlertFiring, "Alert starts firing", corev1.ConditionTrue, timeNow)
			                ^
pkg/handlers/webhookreceiver.go:315:20: Error return value of `status.SetStatus` is not checked (errcheck)
			status.SetStatus(oav1alpha1.ConditionAlertResolved, "Alert has not resolved", corev1.ConditionFalse, timeNow)
			                ^
pkg/handlers/webhookreceiver.go:316:20: Error return value of `status.SetStatus` is not checked (errcheck)
			status.SetStatus(oav1alpha1.ConditionServiceLogSent, "Service log sent for firing alert", corev1.ConditionTrue, timeNow)
			                ^
pkg/k8s/client.go:25:15: Error return value is not checked (errcheck)
	addKnownTypes(scheme)
	             ^
pkg/cli/serve/serve.go:103:18: Error return value of `viper.BindPFlags` is not checked (errcheck)
	viper.BindPFlags(cmd.Flags())
	                ^
pkg/cli/serve/serve.go:142:24: Error return value of `http.ListenAndServe` is not checked (errcheck)
	go http.ListenAndServe(":"+strconv.Itoa(metricsPort), rMetrics)
	                      ^
make: *** [Makefile:123: lint] Error 1
```
Not all errors are related to the metrics (namely `viper.BindPFlags`)

pkg/metrics/metrics.go
bmeng: It will cause the cycle import, where the metrics package imports the handlers package, and handlers package imports the metrics package at the same time. Which is not allowed in golang.

The only use of handlers in metrics.go is the name of those two paths.

mrbarge: Ohhh, right, I get you. Maybe we can break out all the constants into a dedicated `consts` package like we do in ocm-agent-operator. https://github.com/openshift/ocm-agent-operator/tree/master/pkg/consts

pkg/handlers/webhookreceiver.go
bmeng: Another solution is we just declare the variable `metricRequestFailure` and `metricResponseFailure` as `MetricRequestFailure` and `MetricResponseFailure`.

Then the reset func will be
```
func ResetMetric(m *prometheus.GaugeVec) {
	m.Reset()
}
```

And we can call it from the webhookreceiver.go with 
```
	metrics.ResetMetric(metrics.MetricRequestFailure)

```
WDYT?

pkg/metrics/metrics.go
bmeng: yeah, put all the consts into a dedicated package is a good idea

bmeng:
thanks @T0MASD, that's interesting, I am wondering why the IDE cannot detect the problems, like the `error return was not checked`

pkg/handlers/webhookreceiver.go
mrbarge: Ah that also looks good

T0MASD (commented):
left couple of comments

pkg/metrics/metrics.go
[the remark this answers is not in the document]
bmeng: they are already constants? I think you were checking the old commit?

pkg/metrics/metrics.go
```diff
@@ -91,15 +97,35 @@ func PrometheusMiddleware(ph http.Handler) http.Handler {
 }
 
 // SetResponseMetricFailure sets the metric when the call from ocm agent got failed
```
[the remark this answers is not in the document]
bmeng: thanks! updated

pkg/metrics/metrics.go
```diff
@@ -91,15 +97,35 @@ func PrometheusMiddleware(ph http.Handler) http.Handler {
...
-func SetResponseMetricFailure(endpoint string) {
+func SetResponseMetricFailure(service string) {
 	metricResponseFailure.With(prometheus.Labels{
-		"service": endpoint,
+		"ocm_service": service,
 	}).Set(float64(1))
 }
 
 // SetRequestMetricFailure sets the metric when the call to ocm agent got failed
```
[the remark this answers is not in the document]
bmeng: thanks, updated based on your suggestion

pkg/consts/consts.go
```diff
@@ -0,0 +1,20 @@
...
+	MetricsPath = "/metrics"
+	// Ready probe path for OCM Agent web service
+	ReadyzPath = "/readyz"
+	// Live probe path for OCM Agent web service
+	LivezPath = "/livez"
+	// Alertmanger webhook receiver path
+	WebhookReceiverPath = "/alertmanager-receiver"
+
+	// Service name for the sending service logs
+	ServiceLogServiceName = "SREManualAction"
```
ravitri, line 19: For an automated Service Log sent via OCM Agent, should the value still be `SREManualAction`? Not sure if there are other values possible for this or not from OCM side.

bmeng: yeah, I have the same concern, but I did not change it (just moved it from other place to here.)

One thing to think together. The `osdctl servicelog list` is hardcoded to search SLs with the service_name = SREManualAction 
https://github.com/openshift/osdctl/blob/e152fe0698d1ea62c9094abdb48c9021e3d1cde1/cmd/servicelog/list.go#L94

mrbarge: This will probably eventually change as a result of the SDB ticket. It's fine to stay as is for now.

ravitri: Ack.. Resolving comment for now

mrbarge:
Putting a hold back on until CI is healthy..
/hold

mrbarge:
Tested it, working well. From my perspective this is good.
Feel free to unhold when you've got enough reviews and CI is healthy again.
/lgtm

pkg/metrics/metrics.go
```diff
@@ -27,30 +27,37 @@ var (
...
 
-	metricRequestFailure = prometheus.NewGaugeVec(
+	MetricRequestFailure = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "ocm_agent_request_failure",
 			Help: "Indicates that OCM Agent could not successfully process a request",
 		}, []string{"path"})
 
-	metricResponseFailure = prometheus.NewGaugeVec(
+	MetricResponseFailure = prometheus.NewGaugeVec(
```
T0MASD, line 36: just checking if this var name needs to be capitalized, since there are other var names that are not like `metricServiceLogSent` or `metricRequestsTotal`

pkg/metrics/metrics.go
```diff
@@ -27,30 +27,37 @@ var (
...
-		}, []string{"service"})
+		}, []string{"ocm_service"})
+
+	metricServiceLogSent = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "ocm_agent_service_log_sent",
+			Help: "A count of total service log sent based on managedNotification template",
+		}, []string{"ocm_service", "template", "state"})
 
 	metricsList = []prometheus.Collector{
```
T0MASD, line 48: some var names are capitalized while others not, is this intentional?

T0MASD:
Since `ocm_agent_requests_by_service` metric is to collect all the requests received by the  ~~web service  ~~ ocm agent api endpoints, expecting `path` as a param. And to avoid the confusion with `MetricResponseFailure` which expects  service as a param, would this metric be better renamed as `ocm_agent_requests_total` which will make it in line with `ocm_agent_failed_requests_total`?

pkg/metrics/metrics.go
bmeng: yes, the ones capitalized need to be called in the other packages. Mainly used for reset the metric once the failure condition finished.

bmeng: example: https://github.com/openshift/ocm-agent/pull/29/files#diff-76c4aa4812efc417827a4df668f5be466249c5511c82029c475de955ea854a1dR194

PR #38, opened 2022-04-13
OSD-10354 Add metrics test suite

mrbarge:
### What type of PR is this?
Test

Metrics! Tests! Everybody loves them. This PR adds tests for our metrics to make sure they're being set properly when the middleware handler is called, and when our helper functions are called.

As part of writing these tests I noticed we had a slightly brittle piece of code when getting the gorilla mux route in the middleware handler. If the `mux.CurrentRoute(r)` were to ever return nil (which it _can_) that would cause a panic as it is not checked. So that is now made safer.

[OSD-10354](https://issues.redhat.com//browse/OSD-10354)


mrbarge:
/hold 
as I need to make sure that the slight tweak to the middleware handler hasn't broken anything when OA runs.

mrbarge:
/unhold
as it is testing fine

PR #44, opened 2022-07-12
[OSD-12436]add new metric for total SL count based on the managedNotification status

bmeng:
### What type of PR is this?
_(/feature/)_

Add a new metric which to be exposed to observatorium 

### Which Jira/Github issue(s) this PR fixes
_Fixes #OSD-12436_

### DIFF
--- a/pkg/handlers/webhookreceiver.go
+++ b/pkg/handlers/webhookreceiver.go
@@ -205,11 +205,18 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 	}
 
 	// Update the notification status to indicate a servicelog has been sent
-	err = h.updateNotificationStatus(notification, managedNotifications, firing)
+	m, err := h.updateNotificationStatus(notification, managedNotifications, firing)
 	if err != nil {
 		log.WithFields(log.Fields{LogFieldNotificationName: notification.Name, LogFieldManagedNotification: managedNotifications.Name}).WithError(err).Error("unable to update notification status")
 		return err
 	}
+
+	status, err := m.Status.GetNotificationRecord(notification.Name)
+	if err != nil {
+		return err
+	}
+
+	metrics.SetTotalServiceLogCount(notification.Name, status.ServiceLogSentCount)
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


pkg/handlers/webhookreceiver.go
```diff
@@ -211,6 +211,12 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 		return err
 	}
 
+	status, err := managedNotifications.Status.GetNotificationRecord(notification.Name)
+	if err != nil {
+		return err
+	}
+	metrics.SetTotalServiceLogCount(notification.Name, status.ServiceLogSentCount+1)
```
mrbarge, line 218: I've not tested this at all, so I could be wrong.. but won't `ServiceLogSentCount` will have already been incremented by the earlier `updateNotificationStatus` (which increments it via the `SetNotificationRecord` function)?

mrbarge:
Other than my one question, the only other comment I had is, could you please add the new metric to `docs/metrics.md`?
Other than those things it looks fine to me.

pkg/handlers/webhookreceiver.go
bmeng: I tested again, the value set to metric will be 1 less than the one in the managedNotification status.
I think maybe the value fetched in the previous step was not updated one?

But I think there is one problem with the existing format, I made a small change here.

bmeng:
Document updated

mrbarge:
As noted in our separate discussion, the problem is that `updateNotificationStatus` [0] does not update the managed notification it is supplied as a parameter, nor does it return an updated one, so using it for the subsequent `GetNotificationRecord` will result in an error the very first time an SL is sent for a template

We can fix the way that function works to obtain that behaviour and then don't need to do the +1 correction. Because the other problem with the code as-is without that fix is that the very first time an SL is sent the metric won't be set at all (as `GetNotificationRecord` will fail). It only starts being set the second time.

[0] https://github.com/openshift/ocm-agent/blob/master/pkg/handlers/webhookreceiver.go#L338

pkg/handlers/webhookreceiver.go
mrbarge: As noted in our separate discussion, the problem is that updateNotificationStatus [0] does not update the managed notification it is supplied as a parameter, nor does it return an updated one, so using it for the subsequent GetNotificationRecord will result in an error the very first time an SL is sent for a template

We can fix the way that function works to obtain that behaviour and then don't need to do the +1 correction. Because the other problem with the code as-is without that fix is that the very first time an SL is sent the metric won't be set at all (as GetNotificationRecord will fail). It only starts being set the second time.

[0] https://github.com/openshift/ocm-agent/blob/master/pkg/handlers/webhookreceiver.go#L338

bmeng:
Thanks @mrbarge updated based on your suggestion.

mrbarge:
Thanks! Tested and it's working well.
/lgtm

PR #47, opened 2022-11-07
Add logic to avoid sending SL if the resolveBody is empty

weherdh:
### What type of PR is this?
feature

ocm-agent does not send "resolved" ServiceLogs for ManagedNotifications if it is empty

_Fixes #_ [OSD-13528](https://issues.redhat.com/browse/OSD-13528)

### Special notes for your reviewer:
This change is based on this [PR](https://github.com/openshift/ocm-agent-operator/pull/53) 


pkg/handlers/webhookreceiver.go
```diff
@@ -201,7 +202,9 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 	if firing {
 		metrics.CountServiceLogSent(notification.Name, "firing")
 	} else {
-		metrics.CountServiceLogSent(notification.Name, "resolved")
+		if len(notification.ResolvedDesc) > 0 {
```
mrbarge, line 205: Minor thing - as noted in the OAO PR, there's several occasions here of repeated logic about what constitutes when a resolved SL should be sent, so maybe it could be better captured as a function that's part of the ManagedNotification API which OA can leverage. That way if that logic should ever change (extremely unlikely in this case, but who knows), it need only be changed in one place.

pkg/handlers/webhookreceiver.go
```diff
@@ -390,7 +397,11 @@ func (h *WebhookReceiverHandler) updateNotificationStatus(n *oav1alpha1.Notifica
 					// Update the timestamp for the ServiceLogSent
 					_ = status.SetStatus(oav1alpha1.ConditionAlertFiring, "Alert is not firing", corev1.ConditionFalse, timeNow)
 					_ = status.SetStatus(oav1alpha1.ConditionAlertResolved, "Alert resolved", corev1.ConditionTrue, timeNow)
-					_ = status.SetStatus(oav1alpha1.ConditionServiceLogSent, "Service log sent for alert resolved", corev1.ConditionTrue, timeNow)
+					if len(n.ResolvedDesc) > 0 {
+						_ = status.SetStatus(oav1alpha1.ConditionServiceLogSent, "Service log sent for alert resolved", corev1.ConditionTrue, timeNow)
+					} else {
+						status.ServiceLogSentCount--
```
mrbarge, line 403: Why do we need to decrement this?

bmeng:
What will be the status for that managed notification if the alert resolved but we do not send resolve service log.

pkg/handlers/webhookreceiver.go
weherdh: > Why do we need to decrement this?

This is for the total `serviceLogSentCount`. I found this issue when I was doing the test. Can refer to the logic [here](https://github.com/openshift/ocm-agent-operator/blob/master/api/v1alpha1/managednotification_types.go#L266). After the [update status](https://github.com/openshift/ocm-agent/blob/master/pkg/handlers/webhookreceiver.go#L406), the count for sending the service log will be increased which should be wrong since we haven't sent it for resolved ones. I do not want to change more for OAO code, so figured out the idea here.

weherdh:
> What will be the status for that managed notification if the alert resolved but we do not send resolve service log.

I have shared the results in the ticket of this kind of status.

go.mod
```diff
@@ -11,23 +11,122 @@ require (
...
 
 replace (
-	github.com/openshift/api => github.com/openshift/api v0.0.0-20210910062324-a41d3573a3ba
-	k8s.io/api => k8s.io/api v0.21.1
-	k8s.io/apimachinery => k8s.io/apimachinery v0.21.1
-	k8s.io/client-go => k8s.io/client-go v0.21.1
-	k8s.io/kubectl => k8s.io/kubectl v0.21.1
-	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.10.0
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20220414050251-a83e6f8f1d50
+	github.com/openshift/ocm-agent-operator => github.com/weherdh/ocm-agent-operator v0.0.0-20221111023317-5db49a71c2a0
```
weherdh, line 128: This is a temporary resolution. I will update this after OAO related [PR](https://github.com/openshift/ocm-agent-operator/pull/53) merged

go.mod
```diff
@@ -9,25 +9,123 @@ require (
...
 
 replace (
-	github.com/openshift/api => github.com/openshift/api v0.0.0-20210910062324-a41d3573a3ba
-	k8s.io/api => k8s.io/api v0.21.1
-	k8s.io/apimachinery => k8s.io/apimachinery v0.21.1
-	k8s.io/client-go => k8s.io/client-go v0.21.1
-	k8s.io/kubectl => k8s.io/kubectl v0.21.1
-	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.10.0
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20220414050251-a83e6f8f1d50
+	k8s.io/api => k8s.io/api v0.23.5
```
bmeng, line 128: Do we still need those k8s replaces? They are on the same version with above.

bmeng:
the change lgtm from the handling logic, left one comment for the small part.

go.mod
weherdh: Yeah, this is what I copied from [OAO](https://github.com/openshift/ocm-agent-operator/blob/master/go.mod#L101), let me try to remove it and test.

PR #49, opened 2023-03-16
[OSD-15394] Adding Fleetmode changes to ocm-agent CLI

ravitri:
### What type of PR is this?
_(feature)_

Based on [OSD-15394](https://issues.redhat.com/browse/OSD-15394), the purpose of the PR is to incorporate changes in ocm-agent CLI to support "Fleet Mode" to work for Hypershift clusters. In addition to default mode (OSD/ROSA), the `ocm-agent` CLI will support fleet mode as well.

_Fixes #_ [OSD-15394](https://issues.redhat.com/browse/OSD-15394)

### Special notes for your reviewer:
This PR needs to be synced with changes in PR https://github.com/openshift/ocm-agent-operator/pull/61 as well for integration testing.


pkg/cli/serve/serve.go
```diff
@@ -130,13 +175,54 @@ func (o *serveOptions) Run() error {
...
+		if err != nil {
+			o.logger.WithError(err).Fatal("Can't initialise OCM sdk.Connection client in non-fleet mode")
+			return err
+		}
+		o.logger.Info("Connection with OCM initialised successfully in non-fleet mode")
+	} else {
+		// If fleet mode is enabled, the connection to OCM needs to initiate using client ID and client secret
+		// On the managed cluster, the client ID and secret will be fetched from the secret volume however for
+		// local testing, the client ID and secret can be passed directly as flags for ocm-agent CLI.
+		if o.testFleetMode {
```
weherdh, line 193: Just a tiny thing. We do not actually need the flag of `test-fleet-mode`, we can just verify if we pass the `--ocm-client-id` and `--ocm-client-secret`, if yes, read it instead of reading the mounted secret(otherwise if we just pass `test-fleet-mode` without `--ocm-client-id` and `--ocm-client-secret`, what gonna happen?).

ravitri: Well, that's a valid point. My intent was to mainly "group" the flags related to testing based on a boolean flag. Actually I see I haven't added the condition for it but basically to have the client ID and secret passed to the `serve` command ONLY while testing since the secret would expose the credential. Else for production clusters it should be accessible only via volume mounts of the secret. Although, its unlikely that someone were to run the `ocm-agent` binary manually on any production cluster so we could still keep it simple based on your suggestion.

pkg/cli/serve/serve.go
```diff
@@ -69,23 +86,41 @@ func NewServeCmd() *cobra.Command {
 		Long:    serviceLong,
 		Example: serviceExample,
 		Args:    cobra.OnlyValidArgs,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			mode, _ := cmd.Flags().GetBool(config.FleetMode)
+			if !mode {
+				_ = cmd.MarkFlagRequired(config.AccessToken)
+				_ = cmd.MarkFlagRequired(config.ClusterID)
+			}
+			testMode, _ := cmd.Flags().GetBool(config.TestFleetMode)
```
bmeng, line 95: What it is the test fleet mode?

ravitri: Ok, I think I would need to add some documentation around it. But [comment](https://github.com/openshift/ocm-agent/pull/49#discussion_r1141676656) addresses some parts of the answer too. My thoughts were that `ocm-agent` should never consume the OIDC ID and Secret in fleetmode directly as flags but only while "testing/development" work. So I tied `ocm-agent-client-id` and `ocm-agent-client-secret` as mandatory flags only when the test mode flag is passed.

ravitri: If we just want to test fleet mode, we need the secret volume mounts on the cluster. So as an alternative to have the testing possible locally by passing strings, I added the test mode flag as well. I can simplify it by just keeping the client id and secret flags and removing the bool flag if that reduces confusion. Let me know about same.

ravitri:
I have removed the testmode flag and tried to make it simpler by adding more validations instead.

pkg/cli/serve/serve.go
ravitri: Removed the testmode flag.

pkg/cli/serve/serve.go
ravitri: Have removed the testmode flag

pkg/cli/serve/serve.go
```diff
@@ -130,13 +186,54 @@ func (o *serveOptions) Run() error {
...
-	// Initialize ocm sdk connection client
-	sdkclient, err := ocm.NewConnection().Build(viper.GetString(config.OcmURL),
-		viper.GetString(config.ClusterID),
-		viper.GetString(config.AccessToken))
-	if err != nil {
-		o.logger.WithError(err).Fatal("Can't initialise OCM sdk.Connection client")
-		return err
+	// Depending on whether the FleetMode is enabled or not, we need to initiate the OCM SDK connection accordingly
+	// If fleet mode is not enabled, we will fetch the cluster ID and access token to initiate connection with OCM
+	if !o.fleetMode {
```
bmeng, line 191: why do we need to talk to ocm here?

pkg/cli/serve/serve.go
```diff
@@ -130,13 +186,54 @@ func (o *serveOptions) Run() error {
...
+
+			url, err := os.ReadFile(consts.OCMAgentAccessFleetSecretPathBase + os.Getenv("OCM_AGENT_SECRET_NAME") + "/" + consts.OCMAgentAccessFleetSecretURLKey)
+			ocmAgentURL = string(url)
+			if err != nil {
+				o.logger.WithError(err).Fatal("Can't find value for secret key ", consts.OCMAgentAccessFleetSecretURLKey)
+				return err
+			}
+		}
+
+		sdkclient, err = sdk.NewConnectionBuilder().URL(ocmAgentURL).Client(ocmAgentClientID, ocmAgentClientSecret).Insecure(false).Build()
```
bmeng, line 231: why do we need to initialized the connection here?
we should have the metrics set in the separated webhook receivers. So if the connection with OCM failed, we can know when the webhook receiver call failed as well. And we can also choose to be alerted.

pkg/config/config.go
```diff
@@ -11,4 +11,12 @@ const (
 	Debug string = "debug"
 	// ClusterID represents the ID of the cluster used for OCM notifications
 	ClusterID string = "cluster-id"
+	// FleetMode represents if ocm-agent is going to run in default OSD/ROSA mode or HyperShift mode
+	FleetMode string = "fleet-mode"
+	// TestFleetMode represents if ocm-agent is being run locally simulating fleet-mode run
+	TestFleetMode string = "test-fleet-mode"
```
bmeng, line 17: This is left over?

pkg/cli/serve/serve.go
ravitri: Over here we are not really talking to OCM. We are creating the required structs which would help initiate the connection to OCM later based on the credentials passed which is dependent on the mode `ocm-agent` is running in.

pkg/cli/serve/serve.go
ravitri: Over here as well, the connection to OCM is not actually created. The required structs are referred from OCM SDK to create connection being built and constructed which will depend on the OIDC client ID and secret being passed in the fleet mode.

pkg/config/config.go
ravitri: Oh yes, that's a miss from my end. I will remove it.

pkg/cli/serve/serve.go
```diff
@@ -69,23 +85,53 @@ func NewServeCmd() *cobra.Command {
 		Long:    serviceLong,
 		Example: serviceExample,
 		Args:    cobra.OnlyValidArgs,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			clientID, _ := cmd.Flags().GetString(config.OCMClientID)
+			clientSecret, _ := cmd.Flags().GetString(config.OCMClientSecret)
+			mode, _ := cmd.Flags().GetBool(config.FleetMode)
+
+			// Mark AccessToken and ClusterID as required only in default mode
+			if !mode && clientID == "" && clientSecret == "" {
```
bmeng, line 94: This means we will need to set the flags only when both the values are empty?

ravitri: Yes. It's not a nice solution but want I intended to do is that:
- If running `ocm-agent` in default mode, the only acceptable flags should be `--access-token` and `--cluster-id`.
- If running `ocm-agent` in fleet mode (`--fleet-mode`), the `--ocm-client-id` and `--ocm-client-secret` are optional flags for testing but are only valid in fleet mode. 
- For either of the modes, added validations that there shouldn't be a mix of the flags while testing and if it is then the CLI should report the expected behaviour.

PR #50, opened 2023-03-19
Add additional webhookrhobsreceiver for fleet mode

a7vicky:
### What type of PR is this?
_feature_

This PR add additional **webhookrhobsreceiver** for handling alerts on RHOBS cluster to send SL for actionable alerts.

_Fixes #_ [OSD-15393](https://issues.redhat.com/browse/OSD-15393)


pkg/handlers/helper.go
```diff
@@ -0,0 +1,224 @@
...
+	err = responseChecker(operationId, res.Status(), res.Bytes())
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// SendServiceLog sends a servicelog notification for the given alert in fleetmode
+func (o *ocmsdkclient) SendServiceLogInFleetMode(n *oav1alpha1.FleetNotification, firing bool, clusterID string) error {
```
mrbarge, line 157: We can maybe avoid some code duplication here by unifying the two `SendServiceLog..` functions.

- Introduce/keep the `clusterID` parameter, like this function does, but change the non-RHOBS handler to supply the clusterID to it.
- The `if not firing, then generate resolved SL` logic can remain in there rather than be removed, because the RHOBS handler will only ever call the function when an alert is firing (and if it calls the handler with `firing` set to false then that's a bug in the RHOBS handler)

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -0,0 +1,252 @@
...
+		client.InNamespace(OCMAgentNamespaceName),
+	}
+	// Let's get all ManagedFleetNotifications
+	mfnl := &oav1alpha1.ManagedFleetNotificationList{}
+	err := h.c.List(ctx, mfnl, listOptions...)
+	if err != nil {
+		log.WithError(err).Error("unable to list managedFleetNotifications")
+		return &AMReceiverResponse{Error: err, Status: "unable to list managedFleetNotifications", Code: http.StatusInternalServerError}
+	}
+
```
mrbarge, line 92: Now that we're shifting to a design where there's only ever one ManagedFleetNotification per CR, we can simplify this (and the logic in `processAlert`) quite a lot. 

In the non-RHOBS handler it has to search through every `ManagedNotification` looking for the CR that contains the `ManagedNotification` associated with the alert.

In the RHOBS handler we should be able to hopefully do a direct correlation of `the alert defines a managed notification of X, so just get the CR named X` and not do all the list searching - there'll be no need to. That'll hopefully simplify things a bit!

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -0,0 +1,252 @@
...
+	// Can the alert be mapped to an existing notification definition?
+	fn, mfn, err := getFleetNotification(alert.Labels[AMLabelTemplateName], mfnl)
+	if err != nil {
+		log.WithError(err).WithField(LogFieldAlert, fmt.Sprintf("%+v", alert)).Warning("an alert fired with no associated notification template definition")
+		return err
+	}
+
+	mcID := alert.Labels[AMLabelAlertMCID]
+	hcID := alert.Labels[AMLabelAlertHCID]
```
mrbarge, line 124: It may be worth doing a check to ensure that both of these labels actually exist, and skip if they don't.

I'm also thinking of the cases where the hosted cluster ID label doesn't exist, just in case the fleet mode OA ends up getting alerts that are firing on the management cluster (which should be handled by the non-fleet mode OA running on the cluster)

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -0,0 +1,252 @@
...
+			} else {
+				//  add NotificationRecordByName
+				addNotificationRecordByName(fn.Name, fn.ResendWait, mfnr)
+			}
+		} else {
+			// create ManagedFleetNotificationRecord if not found
+			err := h.createManagedFleetNotificationRecord(mcID)
+			if err != nil {
+				return err
+			}
```
mrbarge, line 196: We'll want to move this logic up the top, otherwise the handler will not send a SL for the first time that an alert arrives in for a MC.

ie.
- IF the MFNR doesn't exist; create it
- Then proceed with the rest of the logic around sending an SL

pkg/handlers/helper.go
mrbarge: Ah, now I see why this was done. The different in types for the Notification passed through. :)  
I'll have a think about if we can still improve here.

pkg/handlers/webhookrhobsreceiver.go
mrbarge: (this is now resolved)

pkg/handlers/webhookrhobsreceiver.go
mrbarge: (This is now resolved)

pkg/handlers/helper.go
mrbarge: (this is now resolved, but feedback on the chosen solution of changing the function signature is welcome)

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// processAlert handles the pre-check verification and sending of a notification for a particular alert
+// and returns an error if that process completed successfully or false otherwise
+func (h *WebhookRHOBSReceiverHandler) processAlert(alert template.Alert, mfn oav1alpha1.ManagedFleetNotification) error {
+	fn := mfn.Spec.FleetNotification
+	mcID := alert.Labels[AMLabelAlertMCID]
+	hcID := alert.Labels[AMLabelAlertHCID]
+	slSent := false
+	for !slSent {
```
bmeng, line 110: why do we need a for loop here?

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -0,0 +1,222 @@
...
+		_, err = mfnr.GetNotificationRecordItem(mcID, fn.Name, hcID)
+		if err != nil {
+			// We mustn't, so create one
+			_, err = mfnr.AddNotificationRecordItem(hcID, nfr)
+			if err != nil {
+				return err
+			}
+		}
+
+		// Has a servicelog already been sent
```
bmeng, line 150: small nit, but it would be more accurate to say `If the service log can be sent`

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -0,0 +1,222 @@
...
+			nfr, err = addNotificationRecordByName(fn.Name, fn.ResendWait, hcID, mfnr)
+			if err != nil {
+				return err
+			}
+		}
+
+		// Check if we already have a notification record for this hosted cluster
+		_, err = mfnr.GetNotificationRecordItem(mcID, fn.Name, hcID)
+		if err != nil {
+			// We mustn't, so create one
```
bmeng, line 143: small nit, what does the `We mustn't` here mean?

pkg/handlers/webhookrhobsreceiver.go
mrbarge: Yeah I don't think a loop is needed at all, we can get rid of that.

pkg/handlers/webhookrhobsreceiver.go
mrbarge: The comment here is describing what question the function call is trying to answer, in that sense I feel it is accurate.  An "if the.." comment would be betterr placed above the subsequent "if" block.

pkg/handlers/webhookrhobsreceiver.go
mrbarge: This has now been addressed (for loop removed)

pkg/handlers/webhookrhobsreceiver.go
mrbarge: I've tried to revise the comments both here and in the subsequent if block to be clearer now.

pkg/handlers/webhookrhobsreceiver.go
mrbarge: Comments here have been updated to be clearer.

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (h *WebhookRHOBSReceiverHandler) processAMReceiver(d AMReceiverData, ctx context.Context) *AMReceiverResponse {
+	log.WithField("AMReceiverData", fmt.Sprintf("%+v", d)).Info("Process alert data")
+
+	// Handle each firing alert
+	for _, alert := range d.Alerts.Firing() {
+		// Can we find a notification template for this alert?
+		templateName := alert.Labels[AMLabelTemplateName]
+		mfn := oav1alpha1.ManagedFleetNotification{}
+		err := h.c.Get(ctx, client.ObjectKey{
```
bmeng, line 78: get mfn by template name seems problematic here.

```
time="2023-03-24T06:59:35Z" level=error msg="unable to locate corresponding notification template" error="managedfleetnotifications.ocmagent.managed.openshift.io \"KubeNodeUnschedulableSRE\" not found"

mrbarge:
I'm going to lgtm this to push it along and get an image created. In my own personal tests this seems to be working. We can address anything else we find in subsequent PRs.

/lgtm

PR #66, opened 2023-11-06
[OSD-19304] - service_log handler for sending requests for a specific cluster to the OCM API

xiaoyu74:
### What type of PR is this?
_(feature/refactor)_

**1. Create a Service Log Handler Using the OCM SDK:**
The `ServiceLogHandler` in `service_logs.go` uses the OCM SDK to manage service logs for clusters. It formats and sends log entries via the OCM API using the SDK's `LogEntry` and `ClusterLogsClient`.
```
request := h.ocm.ServiceLogs().V1().ClusterLogs().Add().Body(logEntry)
```

**2. Refactor the existing SendServiceLog function using the new created service log handler:** 
The existing `SendServiceLog()` function in `helper.go`, which originally interacted directly with the OCM API, is being refactored. Instead of making direct API calls, the function now uses the newly created `ServiceLogsHandler` implementing by the existing service log OCM SDK

**P.S.** 
In the initial SendServiceLog function, the `DocReferences` are included in the `sl` object, which is then converted to JSON and sent in the POST request to the OCM API. 
```
sl := ocm.ServiceLog{
		ServiceName:   consts.ServiceLogServiceName,
		ClusterUUID:   clusterID,
		Summary:       summary,
		InternalOnly:  false,
		Severity:      severity,
		LogType:       logType,
		DocReferences: references,
}
```

However, during my refactoring using existing OCM SDK types,  I found the current OCM SDK doesn't support for adding `references` to `LogEntry` (Also read the [OCM SDK doc](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go@v0.1.331/servicelogs/v1) but no functions/types for adding `references`).

So, I just used an alternative way in the refactored `SendServiceLogs()` function in `helper.go` to convert references to JSON and include them in the description.

### Which Jira/Github issue(s) this PR fixes?
https://issues.redhat.com/browse/OSD-19304

PR #68, opened 2023-11-15
OSD-19689: implementing a place holder feature for notifications

Nikokolas3270:
Place holders are replaced with the alert labels & annotations in the service log to generate

### What type of PR is this?
_feature_

This PR allows to use alert labels & annotations in the service logs sent by `ocm-agent`

_Implements_ [OSD-19689](https://issues.redhat.com//browse/OSD-19689)

- [X] Tested latest changes against a cluster
- [X] Included documentation changes with PR

### DIFF
--- a/pkg/handlers/helper.go
+++ b/pkg/handlers/helper.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 
-	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/prometheus/alertmanager/template"
 	log "github.com/sirupsen/logrus"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -31,8 +30,6 @@ const (
 	LogFieldManagedNotification        = "managed_notification_cr"
 	LogFieldPostServiceLogOpId         = "post_servicelog_operation_id"
 	LogFieldPostServiceLogFailedReason = "post_servicelog_failed_reason"
-	ServiceLogActivePrefix             = "Issue Notification"
-	ServiceLogResolvePrefix            = "Issue Resolution"
 
 	// Header returned in OCM responses
 	HeaderOperationId = "X-Operation-Id"
@@ -50,11 +47,6 @@ type AMReceiverData template.Data
 
 type AMReceiverAlert template.Alert
 
-// OCMClient enables implementation of OCM Client
-type ocmsdkclient struct {
-	ocm *sdk.Connection
-}
-
 type WebhookReceiverHandler struct {
 	c   client.Client
 	ocm OCMClient
@@ -62,12 +54,6 @@ type WebhookReceiverHandler struct {
 
 type OCMResponseBody struct {
 	Reason string `json:"reason"`
-}
-
-func NewOcmClient(ocm *sdk.Connection) OCMClient {
-	return &ocmsdkclient{
-		ocm: ocm,
-	}
 }
 
 // isValidAlert indicates whether the supplied alert is one that warrants being processed for a notification.
--- a/pkg/handlers/service_logs.go
+++ b/pkg/handlers/service_logs.go
@@ -4,35 +4,138 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
 
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	slv1 "github.com/openshift-online/ocm-sdk-go/servicelogs/v1"
 	"github.com/openshift/ocm-agent-operator/api/v1alpha1"
 	"github.com/openshift/ocm-agent/pkg/consts"
-	log "github.com/sirupsen/logrus"
+	"github.com/prometheus/alertmanager/template"
 )
 
-type OCMClient interface {
-	SendServiceLog(summary, firingDesc, resolveDesc, clusterID string, severity v1alpha1.NotificationSeverity, logType string, references []v1alpha1.NotificationReferenceType, firing bool) error
+const (
+	ServiceLogActivePrefix  = "Issue Notification"
+	ServiceLogResolvePrefix = "Issue Resolution"
+)
+
+type ServiceLogBuilder struct {
+	wrappedBuilder *slv1.LogEntryBuilder
+	summary        string
+	firingDesc     string
+	resolveDesc    string
+	references     []v1alpha1.NotificationReferenceType
 }
 
-// ServiceLogsHandler manages service logs via the OCM API.
-type ServiceLogsHandler struct {
-	ocm *sdk.Connection
-}
+type ServiceLog = slv1.LogEntry
 
-// NewServiceLogsHandler creates a new handler for service logs.
-func NewServiceLogsHandler(o *sdk.Connection) *ServiceLogsHandler {
-	log.Debug("Creating new service logs handler")
-	return &ServiceLogsHandler{
-		ocm: o,
+func NewServiceLogBuilder(summary, firingDesc, resolveDesc, clusterUUID string, severity v1alpha1.NotificationSeverity, logType string, references []v1alpha1.NotificationReferenceType) *ServiceLogBuilder {
+	return &ServiceLogBuilder{
+		wrappedBuilder: slv1.NewLogEntry().
+			ServiceName(consts.ServiceLogServiceName).
+			ClusterUUID(clusterUUID).
+			InternalOnly(false).
+			Severity(slv1.Severity(severity)).
+			LogType(slv1.LogType(logType)),
+		summary:     summary,
+		firingDesc:  firingDesc,
+		resolveDesc: resolveDesc,
+		references:  references,
 	}
 }
 
-// PostServiceLog sends a service log for a specific cluster to the OCM API.
-func (h *ServiceLogsHandler) PostServiceLog(logEntry *slv1.LogEntry) error {
+var (
+	slVarRefRe = regexp.MustCompile(`\${[^{}]*}`)
+)
+
+// Replace place holders in the given string with the alert labels and annotations
+func replacePlaceHoldersInString(s string, alert *template.Alert) (string, error) {
+	var err error
+	resolvePlaceHolder := func(placeHolder string) string {
+		if err == nil {
+			key, value, isOk := placeHolder[2:len(placeHolder)-1], "", false
+
+			if value, isOk = alert.Labels[key]; !isOk {
+				if value, isOk = alert.Annotations[key]; !isOk {
+					err = fmt.Errorf("alert has no '%s' label or annotation which could be used to replace place holders in the template", key)
+
+					return placeHolder
+				}
+			}
+			return value
+		}
+
+		return placeHolder
+	}
+
+	return slVarRefRe.ReplaceAllStringFunc(s, resolvePlaceHolder), err
+}
+
+func (b *ServiceLogBuilder) Build(firing bool, alert *template.Alert) (*ServiceLog, error) {
+	var summary, description string
+
+	// Adjust the summary based on whether the alert is firing or resolved.
+	if firing {
+		summary = ServiceLogActivePrefix + ": " + b.summary
+		description = b.firingDesc
+
+		if b.references != nil && len(b.references) > 0 {
+			refs, err := json.Marshal(b.references)
+			if err != nil {
+				return nil, fmt.Errorf("failed to marshal references: %v", err)
+			}
+			description = fmt.Sprintf("%s\nReferences: %s", description, string(refs))
+		}
+	} else {
+		summary = ServiceLogResolvePrefix + ": " + b.summary
+		description = b.resolveDesc
+	}
+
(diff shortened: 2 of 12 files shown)


pkg/ocm/servicelog.go
```diff
@@ -0,0 +1,88 @@
...
+					}
+				}
+				return value
+			}
+
+			return placeHolder
+		}
+
+		return slVarRefRe.ReplaceAllStringFunc(s, resolvePlaceHolder), err
+	}
```
typeid, line 76: **Question:** Is there a good reason to declare this function within `ReplacePlaceHoldersWithAlertLabelsAndAnnotations` and `resolvePlaceHolder` within `replacePlaceHoldersInString`? IMO this is pretty hard to read

pkg/ocm/servicelog.go
```diff
@@ -0,0 +1,88 @@
...
+
+	var err error
+
+	if sl.Summary, err = replacePlaceHoldersInString(sl.Summary); err != nil {
+		return err
+	}
```
typeid, line 82: Would save two lines? :) 

```suggestion
	if sl.Summary, err := replacePlaceHoldersInString(sl.Summary); err != nil {
		return err
	}
```

pkg/ocm/servicelog.go
```diff
@@ -0,0 +1,88 @@
...
+
+var (
+	slVarRefRe = regexp.MustCompile(`\${[^{}]*}`)
+)
+
+func (sl *ServiceLog) ReplacePlaceHoldersWithAlertLabelsAndAnnotations(alert *template.Alert) error {
+	replacePlaceHoldersInString := func(s string) (string, error) {
+		var err error
+		resolvePlaceHolder := func(placeHolder string) string {
+			if err == nil {
```
typeid, line 59: What's the  `err == nil` meaning here? When is it not nil, if I'm not mistaken, it seems to just take the nil declaration from above?

pkg/metrics/metrics.go
```diff
@@ -138,6 +145,14 @@ func CountServiceLogSent(template, state string) {
 	}).Inc()
 }
 
+// CountFailedServiceLogs counts the total number of failed service logs (by notification template)
+func CountFailedServiceLogs(template string) {
```
typeid, line 149: Good work on adding this additional metric :)

pkg/metrics/metrics.go
```diff
@@ -17,7 +17,7 @@ var (
 
 	metricRequestsByService = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: "ocm_agent_requests_by_service",
+			Name: "ocm_agent_requests_by_service", // this is a counter without the _total suffix. What about renaming metric to ocm_agent_requests_by_service_total ?
```
bmeng, line 20: Would you like to apply all the changes on the metrics naming suggestion? We can review after that.
But I think if we change the metrics name, we should update all the existing references, like prometheusrules if there is any.
And I remember we have added those metrics to the grafana dashboard, not sure if any update there needed.

pkg/ocm/servicelog.go
Nikokolas3270: `resolvePlaceHolder` function changes the `err` variable in `replacePlaceHoldersInString` function.
For that reason `resolvePlaceHolder` function is nested in in `replacePlaceHoldersInString` function.

I would have preferred to avoid this side effect by the signature of `resolvePlaceHolder` has to be `func(string) string` when calling `slVarRefRe.ReplaceAllStringFunc`.

I initially designed the code by using [`FindStringSubmatchIndex`](https://pkg.go.dev/regexp#Regexp.FindStringSubmatchIndex) in a loop but it was not elegantly possible to deal with recursive variable replacement with that method.

I also though of using [`FindAllStringSubmatchIndex`](https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatchIndex), but the code was then needlessly complicated.

pkg/ocm/servicelog.go
Nikokolas3270: I think that `:=` declare and define/set a variable...
... and `sl.Summary` cannot be declared as it is already declared by being a member of `sl`.

`=` does not have this problem.

pkg/metrics/metrics.go
Nikokolas3270: > Would you like to apply all the changes on the metrics naming suggestion?

Maybe, but not in this pull request as the only purpose of this PR is to add the place holder feature to `ocm-agent`

> But I think if we change the metrics name, we should update all the existing references, like prometheusrules if there is any.
And I remember we have added those metrics to the grafana dashboard, not sure if any update there needed.

Indeed, that's quite some work... also the `PrometheusRule`s and dashboards will probably have to be modified to accommodate both old & new names for a while.
So, to be honest, I don't know if it would be that much of a great idea to rename those metrics.
However I think this should be done in a separate pull request to not clutter and delay this PR which is quite urgent.
But before even submitting a PR for this change, I think it would be best to create a ticket and have the various contributors (team leads, team members, func lead) debate over the necessity of this change. @bmeng : WDYT?
My comments about the names of those metrics was more to open the debate :) ... and I can remove them if we create a ticket to track this renaming question.

pkg/ocm/servicelog.go
typeid: > resolvePlaceHolder function changes the err variable in replacePlaceHoldersInString function.
For that reason resolvePlaceHolder function is nested in in replacePlaceHoldersInString function.

This part makes sense. But can't we move `replacePlaceHoldersInString` out of `ReplacePlaceHoldersWithAlertLabelsAndAnnotations`?

pkg/ocm/servicelog.go
typeid: `:=` works here as well, as it works if only one of the variables needs a fresh initialization. It would only initialize `err` and replace `sl.Summary`

pkg/metrics/metrics.go
bmeng: +1 to put them into separated PR if required
Feel free to create a card to discuss with the owners firstly I would suggest

pkg/ocm/servicelog.go
Nikokolas3270: Technically we could, but `replacePlaceHoldersInString` uses the `alert` parameter passed to `ReplacePlaceHoldersWithAlertLabelsAndAnnotations`; if I move it out of it:
-  I will have to add some `alert` parameter to `replacePlaceHoldersInString`
- By doing so I may have to rename the method to tell that the place holders are replaced with the labels & annotations taken from the new `alert` parameter... while this was implicit with the previous nesting. Method name could become `replacePlaceHoldersWithAlertLabelsAndAnnotationsInString`... this is long, not sure this is suitable.

pkg/ocm/servicelog.go
Nikokolas3270: `:=` only works for variables, not for members:
https://go.dev/play/p/5yJGO_eVEZ8

typeid: TIL. Thanks! :)

pkg/handlers/helper.go
```diff
@@ -59,7 +55,7 @@ type AMReceiverAlert template.Alert
 //
 //go:generate mockgen -destination=mocks/helper.go -package=mocks github.com/openshift/ocm-agent/pkg/handlers OCMClient
 type OCMClient interface {
-	SendServiceLog(summary, firingDesc, resolveDesc, clusterID string, severity v1alpha1.NotificationSeverity, logType string, references []v1alpha1.NotificationReferenceType, firing bool) error
+	SendServiceLog(sl *ocm.ServiceLog) error
```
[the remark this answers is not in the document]
Nikokolas3270: Hello, I am not sure of the urgency of this change (i.e. wether or not we could wait on PR https://github.com/openshift/ocm-agent/pull/66), raising the question to the epic owner.

pkg/metrics/metrics.go
Nikokolas3270: @bmeng: I created a separate ticket for that: [OSD-19707](https://issues.redhat.com/browse/OSD-19707) :-)

Nikokolas3270 (commented):
@bmeng : I removed the comments on the metrics names.

bmeng:
@Nikokolas3270 The #66 merged, please help rebase here

Nikokolas3270:
@bmeng: as discussed, this PR has been adapted to take into account the merge of [PR#66](https://github.com/openshift/ocm-agent/pull/66) which was about changing the way we sent SL to OCM.

I also took the liberty of fixing the 2 following problems I discovered with PR#66:
- Service log description was suffixed with `References: null` is case of `nil` references; that's no longer the case.
- `OCMClient` interface mock was no longer generated. I added back the comment considered by `go generate` in order to be able to generate the mock gain when running `make mockgen`

Finally remark that I also removed the `ServiceLogsHandler` type. Indeed this type was too close to `OCMClient`; the only additional thing this type was doing on top of `OCMClient` was to format the service log; this is now outsourced in a `ServiceLogBuilder` class which is (indirectly) unit tested (c.c. @xiaoyu74 the author of PR#66).

xiaoyu74:
> @bmeng: as discussed, this PR has been adapted to take into account the merge of [PR#66](https://github.com/openshift/ocm-agent/pull/66) which was about changing the way we sent SL to OCM.
> 
> I also took the liberty of fixing the 2 following problems I discovered with PR#66:
> 
> * Service log description was suffixed with `References: null` is case of `nil` references; that's no longer the case.
> * `OCMClient` interface mock was no longer generated. I added back the comment considered by `go generate` in order to be able to generate the mock gain when running `make mockgen`
> 
> Finally remark that I also removed the `ServiceLogsHandler` type. Indeed this type was too close to `OCMClient`; the only additional thing this type was doing on top of `OCMClient` was to format the service log; this is now outsourced in a `ServiceLogBuilder` class which is (indirectly) unit tested (c.c. @xiaoyu74 the author of PR#66).

@Nikokolas3270 - Great work for refactoring my previous PR#66 to separate the log construction from the sending logic with `ServiceLogBuilder`.  It's more clear and modularize.  Thanks.

Just one question why we introduced the function `NewTestServiceLog` considering you've created `ServiceLogBuilder` struct to build service log entries ?

Nikokolas3270:
> Just one question why we introduced the function `NewTestServiceLog` considering you've created `ServiceLogBuilder` struct to build service log entries ?

Hello @xiaoyu74, thanks for your feedback. I considered (for a while) re-using the `ServiceLogBuilder` to create the expected `ServiceLog`, but by doing so:
- I wouldn't have fully tested the `ServiceLogBuilder` as the got & expected `ServiceLog`s would have gone through some shared code to be created.
- Also a new `Build` method (i.e. `BuildTest`) would have needed to be added to `ServiceLogBuilder`. Indeed when creating the expected service log, I know the exact expected summary & description. There should be no choosing between firing & resolving there.
- Similarly a `NewTestServiceLogBuilder` constructor would have been needed. Indeed for the expected service log, there is no need to have an `firingDesc` and a `resolveDesc`.
- All in all this would have cluttered `ServiceLogBuilder` and more code would have been needed.

pkg/handlers/service_logs.go
```diff
@@ -4,88 +4,157 @@ import (
...
+	var err error
+	resolvePlaceHolder := func(placeHolder string) string {
+		if err == nil {
+			key, value, isOk := placeHolder[2:len(placeHolder)-1], "", false
+
+			if value, isOk = alert.Labels[key]; !isOk {
+				if value, isOk = alert.Annotations[key]; !isOk {
+					err = fmt.Errorf("alert has no '%s' label or annotation which could be used to replace place holders in the template", key)
+
+					return placeHolder
```
ravitri, line 60: Can a test be added for this block too? Also, if am reading it right, did you mean `return err` on line 60?

ravitri: Ah ok.. Got it.. :+1:

pkg/handlers/service_logs.go
```diff
@@ -4,88 +4,157 @@ import (
...
-// NewServiceLogsHandler creates a new handler for service logs.
-func NewServiceLogsHandler(o *sdk.Connection) *ServiceLogsHandler {
-	log.Debug("Creating new service logs handler")
-	return &ServiceLogsHandler{
-		ocm: o,
+var (
+	slVarRefRe = regexp.MustCompile(`\${[^{}]*}`)
+)
+
+func replacePlaceHoldersInString(s string, alert *template.Alert) (string, error) {
```
bmeng, line 50: Is it only me who feels the function is hard to understand?...

Nikokolas3270: This function is indeed a bit hard to understand.
I could add some comment telling that the place holders in `s` are replaced with the given `alert` labels & annotations.

The body of the function is also a bit hard to read because of the inner function but the code is actually much shorter that way. Explicitly looping on the matched token is actually clumsier; especially if we want to avoid recursive replacements (and we want that).

PR #69, opened 2023-11-21
OSD-19232 add cluster handler to get cluster object

feichashao:
### What type of PR is this?
Feature

This PR implements a cluster handler which can get the `*cmv1.Cluster` object from OCM.
- It adds an endpoint /my_cluster . `GET /my_cluster` will return the `Cluster` object of the current cluster.
- Add tests for the cluster handler.
- Move some existing functions from `upgrade_policies.go` and `upgrade_policies_test.go` to `handlers_common.go` and `handlers_test.go` for reuse.

https://issues.redhat.com/browse/OSD-19232

_Fixes #_

### DIFF
--- a/pkg/cli/serve/serve.go
+++ b/pkg/cli/serve/serve.go
@@ -291,6 +291,9 @@ func (o *serveOptions) Run() error {
 			r.HandleFunc("/upgrade_policies", upgradePolicyHandler.ServeUpgradePolicyList)
 			r.HandleFunc("/upgrade_policies/{upgrade_policy_id}", upgradePolicyHandler.ServeUpgradePolicyGet)
 			r.HandleFunc("/upgrade_policies/{upgrade_policy_id}/state", upgradePolicyHandler.ServeUpgradePolicyState)
+			o.logger.Info("Initialising Cluster handlers")
+			clusterHandler := handlers.NewClusterHandler(sdkclient, internalID)
+			r.HandleFunc("/my_cluster", clusterHandler.ServeClusterGet)
 		}
 	}
 
--- a/pkg/cli/serve/serve.go
+++ b/pkg/cli/serve/serve.go
@@ -293,7 +293,7 @@ func (o *serveOptions) Run() error {
 			r.HandleFunc("/upgrade_policies/{upgrade_policy_id}/state", upgradePolicyHandler.ServeUpgradePolicyState)
 			o.logger.Info("Initialising Cluster handlers")
 			clusterHandler := handlers.NewClusterHandler(sdkclient, internalID)
-			r.HandleFunc("/my_cluster", clusterHandler.ServeClusterGet)
+			r.HandleFunc("/", clusterHandler.ServeClusterGet)
 		}
 	}
 
--- a/pkg/handlers/cluster.go
+++ b/pkg/handlers/cluster.go
@@ -1,0 +1,57 @@
+package handlers
+
+import (
+	"net/http"
+
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	log "github.com/sirupsen/logrus"
+)
+
+type ClusterHandler struct {
+	ocm       *sdk.Connection
+	clusterId string
+}
+
+// For /api/clusters_mgmt/v1/clusters/{cluster_id}
+// https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id_
+func NewClusterHandler(o *sdk.Connection, clusterId string) *ClusterHandler {
+	log.Debug("Creating new cluster object Handler")
+	return &ClusterHandler{
+		ocm:       o,
+		clusterId: clusterId,
+	}
+}
+
+// https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go@v0.1.382/clustersmgmt/v1#Cluster
+func (g *ClusterHandler) GetCluster(clusterID string) (*cmv1.Cluster, error) {
+	log.Debugf("Sending get cluster object request to OCM API: %s", clusterID)
+	request := g.ocm.ClustersMgmt().V1().Clusters().Cluster(clusterID)
+	resp, err := request.Get().Send()
+	if err != nil {
+		return nil, err
+	}
+	return resp.Body(), nil
+}
+
+// Proxies to
+// https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id_
+func (g *ClusterHandler) ServeClusterGet(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case "GET":
+		cluster, err := g.GetCluster(g.clusterId)
+		if err != nil {
+			errorMessageResponse(err, w)
+			return
+		}
+
+		err = cmv1.MarshalCluster(cluster, w)
+		if err != nil {
+			errorMessageResponse(err, w)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+	default:
+		invalidRequestVerbResponse(r.Method, w)
+	}
+}
--- a/pkg/handlers/cluster.go
+++ b/pkg/handlers/cluster.go
@@ -24,9 +24,9 @@ func NewClusterHandler(o *sdk.Connection, clusterId string) *ClusterHandler {
 }
 
 // https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go@v0.1.382/clustersmgmt/v1#Cluster
-func (g *ClusterHandler) GetCluster(clusterID string) (*cmv1.Cluster, error) {
-	log.Debugf("Sending get cluster object request to OCM API: %s", clusterID)
-	request := g.ocm.ClustersMgmt().V1().Clusters().Cluster(clusterID)
+func (g *ClusterHandler) GetCluster() (*cmv1.Cluster, error) {
+	log.Debugf("Sending get cluster object request to OCM API: %s", g.clusterId)
+	request := g.ocm.ClustersMgmt().V1().Clusters().Cluster(g.clusterId)
 	resp, err := request.Get().Send()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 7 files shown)


pkg/handlers/cluster.go
```diff
@@ -0,0 +1,57 @@
...
+func NewClusterHandler(o *sdk.Connection, clusterId string) *ClusterHandler {
+	log.Debug("Creating new cluster object Handler")
+	return &ClusterHandler{
+		ocm:       o,
+		clusterId: clusterId,
+	}
+}
+
+// https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go@v0.1.382/clustersmgmt/v1#Cluster
+func (g *ClusterHandler) GetCluster(clusterID string) (*cmv1.Cluster, error) {
```
xiaoyu74, line 27: Looks the `ClusterHandler` already contains a `clusterId` field. Is there a specific reason for having a different `cluster ID` to be passed in, the method could use the `clusterId` field directly?
```suggestion
func (g *ClusterHandler) GetCluster() (*cmv1.Cluster, error) {
	log.Debugf("Sending get cluster object request to OCM API: %s", g.clusterId)
	request := g.ocm.ClustersMgmt().V1().Clusters().Cluster(g.clusterId)
```

pkg/handlers/cluster.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return resp.Body(), nil
+}
+
+// Proxies to
+// https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id_
+func (g *ClusterHandler) ServeClusterGet(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case "GET":
+		cluster, err := g.GetCluster(g.clusterId)
```
xiaoyu74, line 42: Also, update the function here accordingly. 
```suggestion
		cluster, err := g.GetCluster()
```

pkg/handlers/cluster.go
```diff
@@ -0,0 +1,57 @@
...
+			errorMessageResponse(err, w)
+			return
+		}
+
+		err = cmv1.MarshalCluster(cluster, w)
+		if err != nil {
+			errorMessageResponse(err, w)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
```
hectorakemp, line 53: @feichashao - we'll need to move this up above `MarshalCluster` like so: 
```
		w.Header().Set("Content-Type", "application/json")
		err = cmv1.MarshalCluster(cluster, w)
```
Otherwise the header won't be set (I figured this out while testing MUO<-> OCM communication and I was wondering why it wasn't serializing!

pkg/cli/serve/serve.go
```diff
@@ -291,6 +291,9 @@ func (o *serveOptions) Run() error {
 			r.HandleFunc("/upgrade_policies", upgradePolicyHandler.ServeUpgradePolicyList)
 			r.HandleFunc("/upgrade_policies/{upgrade_policy_id}", upgradePolicyHandler.ServeUpgradePolicyGet)
 			r.HandleFunc("/upgrade_policies/{upgrade_policy_id}/state", upgradePolicyHandler.ServeUpgradePolicyState)
+			o.logger.Info("Initialising Cluster handlers")
+			clusterHandler := handlers.NewClusterHandler(sdkclient, internalID)
+			r.HandleFunc("/my_cluster", clusterHandler.ServeClusterGet)
```
hectorakemp, line 296: I like the idea here, but my personal preference is for `GET /` to purely reference the cluster object itself. The only reason I say this is because the cluster's OCM pull secret does not have permissions to get other clusters. However, maybe it will someday? 
Maybe we can get Hulk opinions here? 
Do we prefer 
```
			r.HandleFunc("/", clusterHandler.ServeClusterGet)
```
or 
```
			r.HandleFunc("/my_cluster", clusterHandler.ServeClusterGet)
```

pkg/handlers/handlers_common.go
```diff
@@ -0,0 +1,18 @@
+package handlers
+
+import (
+	"fmt"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func errorMessageResponse(err error, w http.ResponseWriter) {
```
hectorakemp, line 10: I like this as an idea :+1:, we can do a follow up PR refactoring upgrade policies to use this

hectorakemp (changes requested):
Great work @feichashao - just a few small nits.

pkg/handlers/cluster.go
feichashao: That's true, I will remove the unnecessary parameter.

pkg/handlers/handlers_common.go
feichashao: the upgrade policy is also in the handler package, thus I moved this out from the upgrade policy file so we know it is a shared function :D

pkg/cli/serve/serve.go
feichashao: > the cluster's OCM pull secret does not have permissions to get other clusters
I believe this is true.

> However, maybe it will someday?
That's what I am concerned about. In case the situation changes and we want to get other clusters, the api design may not look nice if we don't want to break the existing api. It may become:
```
"GET /"  -> the current cluster.
"GET /cluster/{cluster-id}" -> other clusters.
```
Thus I am thinking to explicitly specify it is getting `my_cluster`.

hectorakemp: I definitely get the logic, however we've already built upgrade policies to resolve to 
`get /upgrade_policies`. 
To make things consistent, we would probably have to move it to 
`get /my_cluster/upgrade_policies` 
I've started a discussion in team Hulk about this - we can get their advice on how best to proceed. 
https://redhat-internal.slack.com/archives/C01JJKHC0CF/p1700714500668629

hectorakemp:
Can you squash @feichashao ?

hectorakemp (approved):

pkg/cli/serve/serve.go
feichashao: As per conversation offline, we will use `/` to reflect my cluster as the scope for this is only for it's own cluster.

pkg/handlers/cluster.go
feichashao: Fixed, thanks!

pkg/handlers/cluster.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, err
+	}
+	return resp.Body(), nil
+}
+
+// Proxies to
+// https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_clusters__cluster_id_
+func (g *ClusterHandler) ServeClusterGet(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case "GET":
```
xiaoyu74, line 41: @feichashao - Thanks for the update, just an asking, do we anticipate the need to handle other HTTP methods in the future for something like `POST`/`PUT`/`DELETE` or this handler just strictly dealing with `GET` method? If it's the case,  it might be unnecessary to use `switch` statement ?

xiaoyu74:
@feichashao  - Also, I checked out to your branch and run the `ocm-agent` locally against a staging cluster `26livdquu06lqi60virhq2gskhonshkp`. 
```
--- Running ocm-agent locally...
(Keep this terminal open to follow log of ocm-agent. Open new terminal and run test scripts.)
Link: http://localhost:8081
```

I assume the root path ("/") is being handled by the `clusterHandler.ServeClusterGet` method in the ocm-agent
```
o.logger.Info("Initialising Cluster handlers")
clusterHandler := handlers.NewClusterHandler(sdkclient, internalID)
r.HandleFunc("/", clusterHandler.ServeClusterGet)
```

So, I am using curl to the endpoint directly and I expect to get the cluster info but "page not found".
`curl http://localhost:8081/`

Can you please double check or anything I did wrong for testing ?  cc: @hectorakemp

pkg/handlers/cluster.go
feichashao: Hey @xiaoyu74 , yes, in the current scope we will have `GET` only.  The `switch` is to follow the pattern of other existing handlers [example](https://github.com/openshift/ocm-agent/blob/cb233bb07656357b23cd16e05d00b6b2dc6d449c/pkg/handlers/upgrade_policies.go#L95).

feichashao:
> @feichashao - Also, I checked out to your branch, build and run the `ocm-agent` locally against a staging cluster `26livdquu06lqi60virhq2gskhonshkp`.
> 
> ```
> --- Running ocm-agent locally...
> (Keep this terminal open to follow log of ocm-agent. Open new terminal and run test scripts.)
> Link: http://localhost:8081
> ```
> 
> I assume the root path ("/") is being handled by the `clusterHandler.ServeClusterGet` method in the ocm-agent
> 
> ```
> o.logger.Info("Initialising Cluster handlers")
> clusterHandler := handlers.NewClusterHandler(sdkclient, internalID)
> r.HandleFunc("/", clusterHandler.ServeClusterGet)
> ```
> 
> So, I am using curl to the endpoint directly and I expect to get the cluster info but "page not found". `curl http://localhost:8081/`
> 
> Can you please double check or anything I did wrong for testing ? cc: @hectorakemp

@xiaoyu74 Thanks for testing it!  I forgot to mention that we need to set the cluster_mgmt in the `--services` parameters for testing. Thanks @hectorakemp for pointing that out!

pkg/handlers/cluster.go
xiaoyu74: Yeah,  just thought it's not typical to use switch statement for only one condition. Looks I am too nitpicking, all good. :+1:

pkg/handlers/cluster.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"net/http"
+
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	log "github.com/sirupsen/logrus"
+)
+
+type ClusterHandler struct {
```
bmeng, line 11: Can we change this to `ClustersServiceHandler` or `ClustersMgmtHandler` to match the ocm service better?

feichashao: @bmeng thanks for reviewing!   
The OCM endpoint this handler deal with is 
`​/api​/clusters_mgmt​/v1​/clusters​/` and the corresponding object is [Cluster](https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go@v0.1.382/clustersmgmt/v1#Cluster), so I picked `Cluster` to make it consistent as the name because the `UpgradePolicy` handler is dealing with the object `UpgradePolicy`.

Do you think we can keep it `ClusterHandler` or prefer `ClustersServiceHandler` ?

hectorakemp: `ClusterObjectHandler` could work too?

bmeng:
The change has been tested and worked well.
/lgtm
/approve

PR #70, opened 2023-11-29
[OSD-19745] - Increase the replicas of the ocm-agent pod as part of the HA task

xiaoyu74:
### What type of PR is this?
_(feature)_

Increase the replicas of the ocm-agent to make it Highly Available. It interdependent with the [PR-102](https://github.com/openshift/ocm-agent-operator/pull/102)

https://issues.redhat.com/browse/OSD-19745


a7vicky:
I have raised PR [#104](https://github.com/openshift/ocm-agent-operator/pull/104) to prevent breaking change in case someone sets `ocm-agent` replica count to 1 or 0.

PR #71, opened 2023-12-05
Add clusters_mgmt service on all ocm-agent services

hectorakemp:
### What type of PR is this?
_(feature)_

To allow MUO and other OCM agent consumers to use the `clusters_mgmt` endpoint, we need to make sure that its manager, the OCM agent operator, has enabled the new endpoint.
This enables the `/clusters` endpoint for all ROSA/OSD clusters, so that they can begin to use it OCM agent as an endpoint proxy. 
https://github.com/openshift/managed-cluster-config/pull/1948 depends on it. 

_Fixes #_
https://issues.redhat.com/browse/OSD-19765

### DIFF
--- a/hack/00-ocm-agent.selectorsyncset.yaml.tmpl
+++ b/hack/00-ocm-agent.selectorsyncset.yaml.tmpl
@@ -29,6 +29,7 @@ objects:
           ocmBaseUrl: ${OCM_BASE_URL}
           services:
           - service_logs
+          - clusters_mgmt
         ocmAgentImage: ${REGISTRY_IMG}@${IMAGE_DIGEST}
         replicas: 1
         tokenSecret: ocm-access-token


hack/00-ocm-agent.selectorsyncset.yaml.tmpl
```diff
@@ -32,6 +32,42 @@ objects:
...
+      managed.openshift.io/osd: "true"
+    name: ocm-agent-resources
+  spec:
+    clusterDeploymentSelector:
+      matchLabels:
+        api.openshift.com/managed: "true"
+      matchExpressions:
+      - key: ext-hypershift.openshift.io/cluster-type
+        operator: NotIn
+        values: ["management-cluster"]
```
bmeng, line 51: Why should not be management cluster?

hectorakemp:
> We already have one OcmAgent instance, why we need create a new one?

@wanghaoran1988 the idea is that we are using OA as an OCM endpoint proxy now (MUO will communicate with ocm via OCM agent), and if the node that is hosting the agent goes down (i.e, possibly during an upgrade), we won't be able to communicate with OCM to determine upgrade status. See https://issues.redhat.com/browse/SDE-2150

hectorakemp:
@wanghaoran1988 this one? https://github.com/openshift/ocm-agent/pull/70

wanghaoran1988:
> @wanghaoran1988 this one? #70

No, I mean you are creating two OcmAgent named ocm-agent in this selectorsyncset.

PR #72, opened 2023-12-12
[OSD-19808] - Add doc_references as a separated field from description in SL

xiaoyu74:
### What type of PR is this?
_(bug/feature/test/refactor)_

To have `doc_references` as a separated field from description in SL

https://issues.redhat.com/browse/OSD-19808

### Test steps in an actual staging cluster:
https://gist.github.com/xiaoyu74/d99ab74c735bd1f7d360ca5c81045c5f


pkg/handlers/service_logs.go
```diff
@@ -102,24 +94,49 @@ func (b *ServiceLogBuilder) Build(firing bool, alert *template.Alert) (*ServiceL
...
+		Description(description).
+		DocReferences(docReferences...).
+		Build()
+
+	return logEntry, err
 }
 
-func NewTestServiceLog(summary, desc, clusterUUID string, severity v1alpha1.NotificationSeverity, logType string) *ServiceLog {
-	sl, _ := slv1.NewLogEntry().
+func NewTestServiceLog(summary, desc, clusterUUID string, severity v1alpha1.NotificationSeverity, logType string, references []v1alpha1.NotificationReferenceType) *ServiceLog {
```
bmeng, line 114: Should be better to put this function in test.go?

pkg/handlers/webhookreceiver_test.go
```diff
@@ -73,18 +72,13 @@ var _ = Describe("Webhook Handlers", func() {
 		}
 		testAlert = testconst.NewTestAlert(false, false)
 		testAlertResolved = testconst.NewTestAlert(true, false)
-		activeServiceLog = NewTestServiceLog(
-			ServiceLogActivePrefix+": "+testconst.ServiceLogSummary,
-			testconst.ServiceLogActiveDesc+"\nReferences: [\""+slRefs+"\"]",
-			"",
-			testconst.TestNotification.Severity,
-			testconst.TestNotification.LogType)
 		resolvedServiceLog = NewTestServiceLog(
```
bmeng, line 75: Seems we build the ServiceLog but we never checked the values inside?

pkg/handlers/webhookreceiver_test.go
```diff
@@ -315,7 +309,7 @@ var _ = Describe("Webhook Handlers", func() {
 					},
 				}
 				gomock.InOrder(
-					mockOCMClient.EXPECT().SendServiceLog(activeServiceLog).Return(nil),
+					mockOCMClient.EXPECT().SendServiceLog(gomock.Any()).Return(nil),
```
hectorakemp, line 312: Can we get the 'active' service log in here again using `NewTestServiceLog`?

pkg/handlers/webhookrhobsreceiver_test.go
```diff
@@ -77,7 +70,7 @@ var _ = Describe("RHOBS Webhook Handlers", func() {
 							return nil
 						}),
 					// Send the SL
-					mockOCMClient.EXPECT().SendServiceLog(serviceLog).Return(nil),
+					mockOCMClient.EXPECT().SendServiceLog(gomock.Any()).Return(nil),
```
hectorakemp, line 73: Same as above (include body of `NewTestServiceLog`)

go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/openshift/ocm-agent
 
-go 1.19
+go 1.21
```
hectorakemp, line 3: Do we have to bump Go to bump the OCM sdk too?

pkg/handlers/service_logs.go
hectorakemp, line 114: Should this return the new `*slv1.LogEntry` type?

pkg/handlers/webhookreceiver_test.go
```diff
@@ -49,7 +49,6 @@ var _ = Describe("Webhook Handlers", func() {
 		testAlert                   template.Alert
 		testAlertResolved           template.Alert
 		testManagedNotificationList *ocmagentv1alpha1.ManagedNotificationList
-		activeServiceLog            *ServiceLog
```
hectorakemp, line 52: Should these all be the new type?

pkg/handlers/webhookrhobsreceiver_test.go
```diff
@@ -35,7 +35,6 @@ var _ = Describe("RHOBS Webhook Handlers", func() {
 		testMFN          oav1alpha1.ManagedFleetNotification
 		testMFNR         oav1alpha1.ManagedFleetNotificationRecord
 		mockStatusWriter *clientmocks.MockStatusWriter
-		serviceLog       *ServiceLog
```
hectorakemp, line 38: We could use the new `*slv1.LogEntry` type for this?

hectorakemp (changes requested):
Great work @xiaoyu74, just a few small suggestions (mostly around using the new log entry type in tests). Otherwise, looks great!

pkg/handlers/service_logs.go
xiaoyu74: This function was initially created by Nicolas and I didn't want to alter the previous structure/style but focus on updating the logic to separate the `references` from `description` instead.  But yeah, moving `NewTestServiceLog` function to `service_logs.go` is better considering it's solely used for testing purposes.

pkg/handlers/webhookreceiver_test.go
xiaoyu74: Yeah, I might be ~too lazy~ to replace the `activeServiceLog` instance to use `gomock.Any()` in mocking to less strictly validating the specifics of the `ServiceLog` being passed but benefits of flexibility.  

I agree, use the actual activeServiceLog might be better, let me do some updates accordingly. 
cc: @hectorakemp - I think it also related  to your following questions regarding this building ServiceLog instance?

go.mod
xiaoyu74: no, I just upgraded my local go installation to 1.21 and I think it might be the Go toolchain automatically updated the `go` to reflect the new version. Good catch, I will edit it to revert to 1.19 then.

pkg/handlers/webhookrhobsreceiver_test.go
xiaoyu74: Yes, `*ServiceLog` is just the new `*slv1.LogEntry` type.

hectorakemp (approved):

ravitri:
/approve

Thanks a lot Shawn, Hector and Meng Bo !

PR #75, opened 2024-01-09
Osd 20180 adding dependabot to automatically bump the base image

tkong-redhat:
### What type of PR is this?
Repo enhancement: Automatically bump the base image 

Using Dependabot to automatically bump the base image so we can reduce the manual effort for dependency maintenance.

[_Fixes OSD-20180](https://issues.redhat.com/browse/OSD-20180)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "docker"
+    directory: "/build"
+    labels:
+      - "area/dependency"
+      - "ok-to-test"
+    schedule:
+      interval: "weekly"
+    ignore:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "docker"
```
ravitri, line 3: Can we review a similar repo list CAD and it's configuration and have OAO repo as dependency? Eg - https://github.com/openshift/configuration-anomaly-detection/blob/main/.github/dependabot.yml#L6

tkong-redhat: Config file updated to include the OAO go module check

PR #78, opened 2024-01-16
feat(OSD-20284): implement limited support fleetnotification

typeid:
### What type of PR is this?
feature

This PR implements automation to put clusters in and out of limited support based on `managedfleetnotifications`.
By adding the `limited_support="true"` label to an alert that is sent to the OA alertmanager receiver, we can indicate that instead of sending a service log, OA should put the cluster in limited support.
Resolving alerts with the label result in the cluster being removed from limited support.

Further work in this PR:
- renamed `service_log.go` to `ocm.go` as it wraps different ocm functionalities

Please note that this is WIP, and we possibly want to update the `managedfleetnotification` API as well to be in line with this PR.

https://issues.redhat.com/browse/OSD-20284

- [x] Included documentation changes with PR


pkg/cli/serve/serve.go
```diff
@@ -301,15 +295,22 @@ func (o *serveOptions) Run() error {
...
-			o.logger.Info("Initialising UpgradePolicy handlers")
-			upgradePolicyHandler := handlers.NewUpgradePoliciesHandler(sdkclient, internalID)
-			// See https://github.com/gorilla/mux#examples
-			r.HandleFunc("/upgrade_policies", upgradePolicyHandler.ServeUpgradePolicyList)
-			r.HandleFunc("/upgrade_policies/{upgrade_policy_id}", upgradePolicyHandler.ServeUpgradePolicyGet)
-			r.HandleFunc("/upgrade_policies/{upgrade_policy_id}/state", upgradePolicyHandler.ServeUpgradePolicyState)
-			o.logger.Info("Initialising Cluster handlers")
-			clusterHandler := handlers.NewClusterHandler(sdkclient, internalID)
-			r.HandleFunc("/", clusterHandler.ServeClusterGet)
+			if !o.fleetMode {
```
typeid, line 298: I'm not 100% about this, but I don't see how this would ever run in fleet mode? It needs the cluster id flag which is only passed for non fleet mode?

pkg/handlers/ocm.go
```diff
@@ -113,13 +115,16 @@ func (b *ServiceLogBuilder) Build(firing bool, alert *template.Alert) (*ServiceL
...
+	RemoveLimitedSupport(clusterUUID string, lsReasonID string) error
+	GetLimitedSupportReasons(clusterUUID string) ([]*cmv1.LimitedSupportReason, error)
 }
 
 type ocmClientImpl struct {
 	ocmConnection *sdk.Connection
 }
 
-//go:generate mockgen -destination=mocks/service_logs.go -package=mocks github.com/openshift/ocm-agent/pkg/handlers OCMClient
+//go:generate mockgen -destination=mocks/ocm.go -package=mocks github.com/openshift/ocm-agent/pkg/handlers OCMClient
```
typeid, line 127: I renamed this whole thing `ocm.go`, as it was already a wrapper for ocm and I didn't want to split it up in 2.

test/util/ocm.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+## ocm.sh - Library to handle different ocm operations
```
typeid, line 3: I renamed `servicelog.sh` to `ocm.sh` as well, as it gets the LS functionality. I also updated the api endpoint for the `ocm get /api/service_logs` as the old one was deprecated.

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -95,17 +96,71 @@ func (h *WebhookRHOBSReceiverHandler) processAMReceiver(d AMReceiverData, ctx co
...
+			}
+			metrics.IncrementLimitedSupportRemovedCount(fn.Name)
+		}
+	}
+	// Reset the metric if we got correct Response from OCM
+	metrics.ResetMetric(metrics.MetricResponseFailure)
+
+	// TODO(Claudio): should we add some kind of LimitedSupportRemoved state on the FleetNotificationRecordItem?
+	// Currently, we keep no trace of this other than logs.
+	// We could add a counter for sent limited supports.
```
typeid, line 156: This is likely my biggest open question with my PR: currently, FleetNotificationRecordItem only has a counter for sent SLs (explicitly mentions service logs). 
I think we should rename this to `managed_notifications_fired_sent` and `managed_notification_resolves_sent`, that would keep it neutral when it comes to the type of notification (SL/LS).

typeid:
/unhold 
This should be ready for a full review, I expect minor additions will make it in here or into ocm-agent-operator depending on the review.

docs/metrics.md
```diff
@@ -25,6 +25,10 @@ OCM Agent are managed by [OCM Agent operator](https://github.com/openshift/ocm-a
 |ocm_agent_failed_service_logs_total|Counter|A count of service logs which failed to be sent. This includes service logs which failed to be formatted.|
 |ocm_agent_service_log_sent_total|Gauge|A total number of service log being sent based on managedNotification template|
 |ocm_agent_pull_secret_invalid|Gauge|Pull Secret auth token is not valid|
+|ocm_agent_limited_support_sent_total|Counter| Total number of limited support being sent based on fleetNotification template|
+|ocm_agent_limited_support_removed_total|Counter|Total number of limited support removed based on fleetNotification template|
+|ocm_agent_limited_support_send_failure_total|Counter|Total number of failures for limited support posts based on fleetNotification template|
+|ocm_agent_limited_support_removal_failure_total|Counter|Total number of failures for limited support removals based on fleetNotification template|
```
typeid, line 31: I want to create alerts based on those as well. I'll probably create a followup ticket.

pkg/cli/serve/serve.go
hectorakemp: @typeid we do have plans to eventually use OCM agent as an intermediate proxy in fleet mode, too. See this comment from Ravi: https://redhat-internal.slack.com/archives/C01JJKHC0CF/p1700717837365689?thread_ts=1700714500.668629&cid=C01JJKHC0CF. My preference is to keep it as-is, so that we could use it in the future.

pkg/handlers/ocm.go
hectorakemp: @typeid - the way OA has been organised is kind of 'one handler per OCM API endpoint':
```
$ tree /home/hectorkemp/redhat/go/ocm-agent/pkg/handlers
/home/hectorkemp/redhat/go/ocm-agent/pkg/handlers
├── cluster.go
├── cluster_test.go
├── handlers_common.go
├── handlers_suite_test.go
├── handlers_test.go
├── helper.go
...
├── readyz_test.go
├── service_logs.go
├── service_logs_test.go
├── upgrade_policies.go
├── upgrade_policies_test.go
├── webhookreceiver.go
├── webhookreceiver_test.go
├── webhookrhobsreceiver.go
└── webhookrhobsreceiver_test.go
```
Where `cluster.go` corresponds to `/api/clusters_mgmt/v1/clusters`,
`upgrade_policies.go` corresponds to `/api/clusters_mgmt/v1/upgrade_policies`
`service_logs.go` to `/api/service_logs/v1/cluster_logs/{id}`.
To keep this ordering roughly consistent, could you pull your functions out into a `limited_support.go` handler?

pkg/handlers/webhookrhobsreceiver.go
hectorakemp: I don't have an opinion either way on this - probably will defer to @ravitri (as the SME on OA)

hectorakemp: I do like the idea of a generic `managed_notifications_fired_sent` metric, as it anticipates a broader set of future notifications that can be sent to customers.

pkg/handlers/ocm.go
```diff
@@ -152,3 +157,86 @@ func BuildAndSendServiceLog(slBuilder *ServiceLogBuilder, firing bool, alert *te
...
+	// Check the response status code
+	if response.Status() < 200 && response.Status() >= 300 {
+		// Extract error details from the response and return an appropriate error.
+		return nil, fmt.Errorf("unexpected status: %d", response.Status())
+	}
+
+	return response.Items().Slice(), nil
+}
+
+func (o *ocmClientImpl) getInternalID(clusterUUID string) (string, error) {
```
hectorakemp, line 222: You can use the existing function from here: https://github.com/openshift/ocm-agent/blob/master/pkg/ocm/connection.go

test/util/ocm.sh
hectorakemp: I think renaming this shell script to `ocm.sh` is definitely appropriate here.

pkg/handlers/ocm.go
hectorakemp: See this recently completed epic (https://issues.redhat.com/browse/SDE-2150) for further explanation on why we do/did this.

hectorakemp (changes requested):
Great work so far @typeid, all of the logic changes look sound and well thought out. Just a few organisational nits that I've pointed out in comments below.

pkg/cli/serve/serve.go
typeid: Confirmed that the above crashloops OCM-agent with the current main/prod deployment on the rhobs cluster:
```
oc logs ocm-agent-staging-hypershift-6c4557d758-8948q -n openshift-ocm-agent-operator
***** Starting with FIPS crypto enabled *****
time="2024-01-18T06:43:36Z" level=info msg="Starting ocm-agent server"
time="2024-01-18T06:43:36Z" level=info msg="Fleet mode configured" FleetMode=true
time="2024-01-18T06:43:36Z" level=info msg="Start listening on metrics port" Port=8383
time="2024-01-18T06:43:36Z" level=info msg="Connection with OCM initialised successfully in fleet mode"
time="2024-01-18T06:43:37Z" level=error msg="Cluster with external id  not found in OCM database."
time="2024-01-18T06:43:37Z" level=fatal msg="OCM Agent failed to fetch internal cluster ID" error="cluster with external id  not found in OCM database"
```

The reason I had added a check for the fleet-mode there was that it would crash if I didn't: there is no cluster-id flag passed to ocm-agent in fleet mode, so IMO this can't work and something is wrong.

typeid: Fixed with https://github.com/openshift/ocm-agent/pull/79

pkg/handlers/webhookrhobsreceiver.go
typeid: For the API: https://github.com/openshift/ocm-agent-operator/pull/106
As noted in that PR, I am unsure what the flow here is, as the api is needed for https://github.com/openshift/ocm-agent/pull/78, but the api also doesn't fit the currently deployed ocm-agent. Do we have some kind of circular dependency here?

pkg/handlers/ocm.go
typeid: Will address same comment below. Thanks!

typeid: While I see what's trying to be done here, it doesn't make sense to me for the alertmanager webhooks. There's no direct 1:1 mapping from webhook to an OCM endpoint so ocm-agent can't be a "reverse proxy" as described in that epic. 

See e.g., I hope this shows the issue :
```
for _, service := range o.services {
			switch service {
			case config.ServiceLogService:
				o.logger.Info("Initialising alertmanager webhook handler in NON-fleet mode")
				webhookReceiverHandler := handlers.NewWebhookReceiverHandler(client, ocmclient)
				r.Path(consts.WebhookReceiverPath).Handler(webhookReceiverHandler)
				r.Use(metrics.PrometheusMiddleware)
			case config.ClustersService:
				o.logger.Info("Initialising UpgradePolicy handlers")
				upgradePolicyHandler := handlers.NewUpgradePoliciesHandler(sdkclient, internalID)
				// See https://github.com/gorilla/mux#examples
				r.HandleFunc("/upgrade_policies", upgradePolicyHandler.ServeUpgradePolicyList)
				r.HandleFunc("/upgrade_policies/{upgrade_policy_id}", upgradePolicyHandler.ServeUpgradePolicyGet)
				r.HandleFunc("/upgrade_policies/{upgrade_policy_id}/state", upgradePolicyHandler.ServeUpgradePolicyState)
				o.logger.Info("Initialising Cluster handlers")
				clusterHandler := handlers.NewClusterHandler(sdkclient, internalID)
				r.HandleFunc("/", clusterHandler.ServeClusterGet)
			}
```

typeid: As discussed, moved `ocm.go` out into `pkg/ocm`.

hectorakemp (approved):

hectorakemp:
@ravitri - @typeid and I had a meeting, and realized that the way that we were doing handlers right now doesn't make sense. We're mixing HTTP request/response logic (i.e, the reverse proxy handlers) with the OCM wrapper logic. 
We decided it makes more sense to wrap all stuff that purely wraps the OCM API/SDK in a `pkg/ocm.go` file, which can then be called by the HTTP handlers. 
I'm going to do a follow up PR that does the same for all of the recent work done for the endpoint proxy feature.

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -95,17 +97,71 @@ func (h *WebhookRHOBSReceiverHandler) processAMReceiver(d AMReceiverData, ctx co
...
 		}
 	}
+
 	return &AMReceiverResponse{Error: nil, Status: "ok", Code: http.StatusOK}
 }
 
-// processAlert handles the pre-check verification and sending of a notification for a particular alert
+// processResolvedAlert handles resolve notifications for a particular alert
+// currently only handles removing limited support
+func (h *WebhookRHOBSReceiverHandler) processResolvedAlert(alert template.Alert, mfn oav1alpha1.ManagedFleetNotification) error {
```
weherdh, line 124: A question here: The alert for the LS is a resolved one? Or the alert for LS can be a firing one? If the alert has been resolved why we still need to put the cluster to limited support? So maybe the name `processLimitSupport` is better?

typeid: The alert for LS can be both firing and resolving:
- if the alert is firing and contains `limited_support = true`, it will be put in limited support
- if the alert is resolving and contains `limited_support = true`, it will be removed from limited support.

I kept the naming of this method neutral to keep it extensible for the future and be in line with the way the standard webhook was implemented. 
This method could handle resolving alerts in a different ways (e.g. `send resolved servicelog` for the non-RHOBS webhooks) in the future. But you are correct, for now this function only handles resolving alerts that contain the `limited_support` label, for the rest it would return `nil`.

PR #82, opened 2024-01-31
Fix(OSD-20662): race condition in fleet mode alert handling

typeid:
### What type of PR is this?
bug

Fixes race condition in fleet mode alert handling. We were not handling conflicts on kube resource updates.

Fixes [OSD-20662](https://issues.redhat.com//browse/OSD-20662)

### DIFF
--- a/pkg/consts/test/test.go
+++ b/pkg/consts/test/test.go
@@ -114,7 +114,7 @@ func NewFleetNotification() ocmagentv1alpha1.FleetNotification {
 func NewManagedFleetNotification(limited_support bool) ocmagentv1alpha1.ManagedFleetNotification {
 	mfn := ocmagentv1alpha1.ManagedFleetNotification{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "test-mfn",
+			Name:      TestNotificationName,
 			Namespace: "openshift-ocm-agent-operator",
 		},
 		Spec: ocmagentv1alpha1.ManagedFleetNotificationSpec{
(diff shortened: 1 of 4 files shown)


xiaoyu74:
@typeid  - Great work to introduce `RetryOnConflict` to address the potential race condition case.  I also got you encapsulated the logic for creating or updating the `ManagedFleetNotificationRecord` into dedicated methods below which is nice:

- `getOrCreateManagedFleetNotificationRecord`
-  `updateManagedFleetNotificationRecord`

From the test cases, it looks focus the functional correctness of handling alerts and managing fleet notification records rather than explicitly validating the effectiveness of the `RetryOnConflict` strategy to mitigating race conditions?

If it's better we can extend the test cases to explicitly simulate conflict errors on operations to verify that the `RetryOnConflict` strategy is invoked as expected since it's the main function the PR introduced for mitigating the race condition right ?

typeid:
@xiaoyu74 Thanks so much for the review!

I introduced an integration test to validate concurrent handling of webhooks, I believe that should already better cover the situation. 
As for unit testing the behavior with `RetryOnConflict`: what are we trying to validate in these tests? I'm usually against testing included packages' functions, but if you believe there's a specific case we want to cover I'm happy to add a test. 

E.g., what I would not see as useful and covered by the included packages' tests: a test that runs `updateManagedFleetNotificationRecord` and mocks a conflict error.

xiaoyu74:
@typeid - Thanks for your clarifications, given the situation that you've introduced integration test validates concurrent handling of webhooks, it might not be strictly necessary to add unit tests focused solely on `RetryOnConflict`. 

Actually, I can give the PR a LGTM but just sharing some of my thoughts below: 

I agree that the function `RetryOnConflict` provided by the external `client-go` lib should have been well-tested. What I was thinking is that, adding the unit tests using the `RetryOnConflict` **is not about testing the function itself** but ensuring the application's logic that integrates with the lib is correct.

For example, the webhookhandler needs to update MFNR, conflicts are expected under certain conditions (e.g high concurrency), the application logic decides to use `RetryOnConflict`, but when and why iit decide that a  retry is necessary? Is the logic based on specific error code, state of the MFNR resource or other conditions? So, the unit test can ensure if the decision making is correct. 

Also, what happens after `RetryOnConflict` resolves the conflict or what if it fails? The application logic will do something right ? (e.g update cache, inform user or roll back), the unit test can check the follow-up action are correctly implemented. 

So, again, the unit test is not test the function itself but to test if the application (webhook handler) logic integrated with the function (`RetryOnConflict`) as expected. Also, unit tests execute more quickly than integration tests, providing us with a faster feedback loop.

typeid:
@xiaoyu74 
> the application logic decides to use RetryOnConflict, but when and why iit decide that a retry is necessary? Is the logic based on specific error code, state of the MFNR resource or other conditions? So, the unit test can ensure if the decision making is correct.

That's precisely the implementation of `RetryOnConflict`, if an error is of type conflict, it retries.

But isn't that exactly what `RetryOnConflict` implements its own tests should cover? It retries when there's a conflict error. This error is not related to our resources. If an error that is not a conflict is returned it will return an error.

>Also, what happens after RetryOnConflict resolves the conflict or what if it fails? The application logic will do something right ?

It would bubble up the error to lead to a failure.

Would a test case where the `getOrCreateManagedFleetNotificationRecord` and `updateManagedFleetNotificationRecord` error, and the error gets bubbled up address your concerns?

ravitri:
Based on slack conversation, we will take up further test improvements in separate ticket dedicated to tests improvement but unblock this PR for now. Thanks Shawn for the reviews so far and Claudio for the integration test and comments. 

/lgtm
/approve

PR #88, opened 2024-02-07
fix(OSD-20857): resend LS only when resolved

typeid:
**Note:** **this is a branch off https://github.com/openshift/ocm-agent/pull/82 - once PR 82 is merged, the diff here will be much smaller.** 

### What type of PR is this?
bug

When alertmanager restarts, all alerts are resent. This PR adds a check to make sure we don't send limited support if it's already active. No changes to service logs.

I also included an E2E test for this.

Fixes [OSD-20857](https://issues.redhat.com//browse/OSD-20857)

### DIFF
--- a/pkg/handlers/webhookrhobsreceiver.go
+++ b/pkg/handlers/webhookrhobsreceiver.go
@@ -193,12 +193,12 @@ func (h *WebhookRHOBSReceiverHandler) processFiringAlert(alert template.Alert, m
 	}
 
 	// Check if a firing notification can be sent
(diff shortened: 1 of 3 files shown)


pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -353,3 +353,39 @@ func (h *WebhookRHOBSReceiverHandler) updateManagedFleetNotificationRecord(alert
 
 	return err
 }
+
+// firingCanBeSent wraps the api's `mfnr.FiringCanBeSent`
```
typeid, line 357: I wrapped this as I did not want to modify the API again. We should ideally have this function here, not in the API.

PR #89, opened 2024-02-14
Fix: CanBeSent does not rely on existing object

typeid:
### What type of PR is this?
bug

**The bug:** 
- with the [latest fix of a nil ptr deref](https://github.com/openshift/ocm-agent-operator/pull/111) in `ocm-agent-operator`, recorditems with a `nil` `lastUpdatedTimestamp` are automatically garbage collected
- we currently need to pre-create recorditems for the `CanBeSent()` function, but the timestamp isn't set

**Result:** the recorditems we create are instantly garbage collected by ocm-agent-operator

**The fix:**

`FiringCanBeSent()` defined in `ocm-agent-operator` (API) requires a `managedfleetnotificationrecord` to exist: https://github.com/openshift/ocm-agent-operator/blob/e0178327b55ced0475e276c147e7b73b7c4cbfe1/api/v1alpha1/managedfleetnotificationrecord_types.go#L145
Due to this, we needed to create it before the check. 

This PR re-implements `firingCanBeSent()` to not rely on the existing record: if no record exists, the notification can automatically be sent. 
With this change, the creation and update of the `managedfleetnotificationrecord / item` was moved into a single function that is called once OA handles a managed fleet notification. This allows creating the recorditem directly with the timestamp in a single `Update()` call.

Note that we can delete the `FiringCanBeSent` in ocm-agent-operator after this PR.

_Fixes #_

### DIFF
--- a/pkg/handlers/webhookrhobsreceiver_test.go
+++ b/pkg/handlers/webhookrhobsreceiver_test.go
@@ -88,10 +88,6 @@ var _ = Describe("RHOBS Webhook Handlers", func() {
 							Expect(mfnr.Name).To(Equal(testconst.TestManagedClusterID))
 							return nil
 						}),
-					// Initially add the MC name to the status as it isn't set yet
-					mockClient.EXPECT().Status().Return(mockStatusWriter),
-					mockStatusWriter.EXPECT().Update(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil),
-
 					// Update status for the handled alert
 					mockClient.EXPECT().Status().Return(mockStatusWriter),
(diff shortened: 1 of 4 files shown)


pkg/handlers/webhookrhobsreceiver_test.go
```diff
@@ -347,6 +333,7 @@ var _ = Describe("RHOBS Webhook Handlers", func() {
 							},
 						},
 					}
+					testMFN.Spec.FleetNotification.ResendWait = 24
```
typeid, line 336: We're now going off the fleetnotification's resendwait, not the one specific for a record.

PR #90, opened 2024-02-15
OSD-20982: add check for the recorditem states to integration tests

typeid:
### What type of PR is this?
test

This PR adds more tests for the previous PR https://github.com/openshift/ocm-agent/pull/89. These tests would have caught the unwanted garbage collection of recorditems before it made it to prod.

_Fixes #_

### DIFF
--- a/pkg/handlers/webhookrhobsreceiver.go
+++ b/pkg/handlers/webhookrhobsreceiver.go
@@ -286,7 +286,7 @@ func (h *WebhookRHOBSReceiverHandler) updateManagedFleetNotificationRecord(alert
 		// Fetch the ManagedFleetNotificationRecord, or create it if it does not already exist
 		mfnr, err := h.getOrCreateManagedFleetNotificationRecord(mcID, hcID, mfn)
 		if err != nil {
-			log.WithFields(log.Fields{LogFieldNotificationRecordName: mfnr.Name}).Infof("getOrCreate of managedfleetnotificationrecord failed: %s. Retrying in case of conflict error", err.Error())
+			log.WithFields(log.Fields{LogFieldNotificationRecordName: mcID}).Infof("getOrCreate of managedfleetnotificationrecord failed: %s. Retrying in case of conflict error", err.Error())
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -286,7 +286,7 @@ func (h *WebhookRHOBSReceiverHandler) updateManagedFleetNotificationRecord(alert
 		// Fetch the ManagedFleetNotificationRecord, or create it if it does not already exist
 		mfnr, err := h.getOrCreateManagedFleetNotificationRecord(mcID, hcID, mfn)
 		if err != nil {
-			log.WithFields(log.Fields{LogFieldNotificationRecordName: mfnr.Name}).Infof("getOrCreate of managedfleetnotificationrecord failed: %s. Retrying in case of conflict error", err.Error())
+			log.WithFields(log.Fields{LogFieldNotificationRecordName: mcID}).Infof("getOrCreate of managedfleetnotificationrecord failed: %s. Retrying in case of conflict error", err.Error())
```
typeid, line 289: Small logging bug I found on the way: mfnr.Name could be nil here, but it is equal to mcID, so we can use that for logging instead.

test/test-fleet-mode-alerts.sh
```diff
@@ -59,62 +61,63 @@ PreCheck
...
+PRE_SL_COUNT=$(get-servicelog-count ${EXTERNAL_ID})
 
 for ((i=1; i<=10; i++)); do
-  post-alert ${ALERT_FILE_FIRING} &
+  post-alert ${TMP_ALERT_FILE} &
 done
-sleep 15 # Wait for everything to be handled
-EXPECTED_COUNT=$((${PRE_LS_COUNT} + 10))
-check-limited-support-count ${EXTERNAL_ID} ${EXPECTED_COUNT}
+sleep 25 # Wait for everything to be handled
```
typeid, line 121: Updated the sleep to be a tick longer to make sure we get no failures sporadically when processing is slow due to retries.

PR #91, opened 2024-02-22
Move OCM SDK handler methods to pkg/ocm.go

hectorakemp:
### What type of PR is this?
_(cleanup)_

- Reorganizes the `upgrade_policies.go` handler. Instead of declaring OCM SDK methods within the handler itself, it moves them to the `pkg/ocm.go` parent package. This allows them to be called by other handler methods cleanly that aren't the upgrade policies handler. It also makes the code a bit nicer to deal with. 

_Fixes #_
https://issues.redhat.com/browse/OSD-20778


pkg/ocm/ocm.go
```diff
@@ -130,6 +137,82 @@ func NewOcmClient(ocmConnection *sdk.Connection) OCMClient {
...
 
+// https://pkg.go.dev/github.com/openshift-online/ocm-sdk-go@v0.1.382/clustersmgmt/v1#Cluster
+func (o *ocmClientImpl) GetCluster(clusterID string) (*cmv1.Cluster, string, error) {
+	log.Debugf("Sending get cluster object request to OCM API: %s", clusterID)
+	request := o.ocmConnection.ClustersMgmt().V1().Clusters().Cluster(clusterID)
+	resp, err := request.Get().Send()
+	if err != nil {
+		return nil, resp.Header().Get(OcmOperationIdHeader), err
+	}
+	return resp.Body(), resp.Header().Get(OcmOperationIdHeader), nil
```
typeid, line 148: **suggestion [repeated for other functions containing ocm requests]:** We should probably check `resp.Status()` here: we could in theory have a HTTP request that didn't result in an error (the request itself worked), but the status could be a 500. 

The other functions here do:
```
	// Check the response status code
	if response.Status() < 200 && response.Status() >= 300 {
		return fmt.Errorf("unexpected status: %d", response.Status())
	}
```

typeid (commented):
Very nice cleanup, it's looking great @hectorakemp! I added one small suggestion, but other than that it looks good to go.

pkg/ocm/ocm.go
hectorakemp: @typeid - this is already handled by the OCM SDK (and returns an error). This holds for all of the HTTP transports methods in the OCM SDK. See here: https://github.com/search?q=repo%3Aopenshift-online%2Focm-sdk-go%20errors.UnmarshalErrorStatus&type=code

hectorakemp: I think that this also means that you can remove these lines from `SendLimitedSupport`, `RemoveLimitedSupport`, etc.

typeid: If I read it correctly, this only handles `result.status >= 400`, what about other status codes?

hectorakemp: Ah, you are indeed correct @typeid. Great pickup :) 

I've added 3xx error handling in `pkg/ocm.go`, some basic tests, as well fixed a type in the conditionals for the service log/limited support functions. 
```
		if resp.Status() < 200 && resp.Status() >= 300 {
```
was never being reached for 3xx responses. I've updated it to 
```
		if resp.Status() < 200 || resp.Status() >= 300 {
```
for all methods in `pkg/ocm.go`.

PR #106, opened 2024-08-16
remove the validating for source field for fleet mode alerts

bmeng:
### What type of PR is this?
_(bug)_

We are validating the alert with label before proceed it with the ocm-agent service log sending.
There was a `source` label when the alerts are handled by the rhobs, but the field is not there when we move the alert handling to the obo stack running on the management cluster.
The existing validator with alert label _id and _mc_id should be able to indicate the alert is valid.

Fixes #[OSD-25042](https://issues.redhat.com//browse/OSD-25042)

### DIFF
--- a/pkg/handlers/helper.go
+++ b/pkg/handlers/helper.go
@@ -5,10 +5,11 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/openshift/ocm-agent/pkg/ocm"
 	"github.com/prometheus/alertmanager/template"
 	log "github.com/sirupsen/logrus"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/ocm-agent/pkg/ocm"
 
 	_ "github.com/golang/mock/mockgen/model"
 )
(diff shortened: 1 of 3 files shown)


pkg/handlers/helper.go
```diff
@@ -5,23 +5,24 @@ import (
...
 	AMLabelTemplateName        = "managed_notification_template"
 	AMLabelManagedNotification = "send_managed_notification"
-	AMLabelAlertSourceName     = "source"
-	AMLabelAlertSourceHCP      = "HCP"
-	AMLabelAlertSourceDP       = "DP"
 	AMLabelAlertMCID           = "_mc_id"
 	AMLabelAlertHCID           = "_id"
+	//AMLabelAlertSourceName     = "source"
+	//AMLabelAlertSourceHCP      = "HCP"
+	//AMLabelAlertSourceDP       = "DP"
```
ravitri, line 25: These commented constants can be removed instead?

bmeng: yeah, good to be removed

typeid (approved):
/lgtm
/approve

PR #108, opened 2024-08-28
retry send SL notification

a7vicky:
### What type of PR is this?
_(refactor)_

Retry if it fails to send SL.

_Fixes #_[OSD-20514](https://issues.redhat.com/browse/OSD-20514)


pkg/handlers/webhookreceiver_test.go
ravitri, line 1: Added PR https://github.com/a7vicky/ocm-agent/pull/8 for small fix to the tests

pkg/handlers/webhookreceiver.go
```diff
@@ -148,14 +148,27 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
...
+	var retrycount int = 0
+	for {
+		slerr := ocm.BuildAndSendServiceLog(ocm.NewServiceLogBuilder(notification.Summary, notification.ActiveDesc, notification.ResolvedDesc, viper.GetString(config.ExternalClusterID), notification.Severity, notification.LogType, notification.References), firing, &alert, h.ocm)
+		if slerr == nil {
+			break
+		} else {
+			log.WithError(err).WithFields(log.Fields{LogFieldNotificationName: notification.Name, LogFieldIsFiring: true}).Error("unable to send a notification")
+			retrycount++
+			time.Sleep(10*time.Second + 5*time.Second*time.Duration(time.Duration(math.Pow(2, float64(retrycount))).Seconds()))
+			if retrycount == 7 {
```
ravitri, line 161: Overall, I think having time based retry limit would be better than count based. Initially a static timeout would be fine too (Eg. 5 minutes). We can still have exponential retries in that interval. Eventually the timeout could be configurable as well and can be easier for us to manage in tests than always having to wait for 7 retries even in tests.

a7vicky:
tide/merge-method-squash

pkg/handlers/webhookreceiver.go
```diff
@@ -148,15 +146,24 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 	}
 	// Send the servicelog for the alert
 	log.WithFields(log.Fields{LogFieldNotificationName: notification.Name}).Info("will send servicelog for notification")
-	err = ocm.BuildAndSendServiceLog(
-		ocm.NewServiceLogBuilder(notification.Summary, notification.ActiveDesc, notification.ResolvedDesc, viper.GetString(config.ExternalClusterID), notification.Severity, notification.LogType, notification.References),
-		firing, &alert, h.ocm)
-	if err != nil {
-		log.WithError(err).WithFields(log.Fields{LogFieldNotificationName: notification.Name, LogFieldIsFiring: true}).Error("unable to send a notification")
+
+	var attempts int = 5
```
chamalabey, line 150: WDYT about including the number of attempts and sleep duration in the ocm-agent ConfigMap?

a7vicky: @chamalabey Thanks for the review. This is something I do think that these parameters shouldn't be exposed externally. I couldn't of any scenario where we want this to be configurable for different use cases?

chamalabey: My thinking was not hard code the values in the code and have them in a configmap so they can be configured without code changes. If you think this is not going to be changed for any reason, it is okay to hard code it.

a7vicky:
Closed this MR with new MR #111

PR #111, opened 2024-09-22
[OSD-20514] Retry send SL notification

a7vicky:
### What type of PR is this?
_(feature)_

Retry if it fails to send SL.

_Fixes #_[OSD-20514](https://issues.redhat.com/browse/OSD-20514)


pkg/handlers/webhookreceiver.go
```diff
@@ -146,17 +145,34 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 		// This is not an error state
 		return nil
 	}
+
+	var attempts int = 5
+	var sleep time.Duration = 30
```
tkong-redhat, line 150: Was Duration unit is nanoseconds? Just wonder if 30 nanoseconds is longer enough for retry

a7vicky: Thanks @tkong-redhat for the review. I have updated this in my latest commit PTAL

pkg/httpchecker/reattempt.go
```diff
@@ -0,0 +1,35 @@
...
+	"time"
+)
+
+func init() {
+	rand.New(rand.NewSource(time.Now().UnixNano())) //nolint:gosec
+}
+
+type stop struct {
+	error
+}
```
devppratik, line 14: Does this need to a struct separately if we only have error here?

pkg/httpchecker/urlavailabilitychecker.go
```diff
@@ -0,0 +1,41 @@
...
+func (c *UrlHTTPChecker) UrlAvailabilityCheck(url string) error {
+	resp, err := c.Client.Get(url)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return nil
+	}
```
devppratik, line 39: Can we improve this check?
Maybe using somethine like 
```
if resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices {
    return nil
}
```
OR 
```
if resp.StatusCode/100 == 2 {
    return nil
}
```

devppratik (commented):
LGTM on the code. Just a few non blocking review

pkg/httpchecker/urlavailabilitychecker.go
a7vicky: Thanks @devppratik for the review. I have updated in my latest commit

pkg/httpchecker/reattempt.go
a7vicky: Yes this is not needed but I have keep it consistent as same and ref from MUO repo

PR #113, opened 2024-09-30
Updating base image to match boilerplate for Go 1.22

ravitri:
### What type of PR is this?
_(bug)_

Raising this PR to fix the base image for ocm-agent to not use Centos base image but aligning it to use boilerplate image. Without this fix, ocm-agent has failure like:
```shell
2024-09-30T04:12:24.317367291+00:00 stderr F ocm-agent: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ocm-agent)
2024-09-30T04:12:24.317367291+00:00 stderr F ocm-agent: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ocm-agent)
```

Thanks @tkong-redhat and @a7vicky to figure out the issue :+1:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ docs:
 # Installed using instructions from: https://golangci-lint.run/usage/install/#linux-and-windows
 getlint:
 	@mkdir -p $(GOPATH)/bin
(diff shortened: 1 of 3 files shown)


typeid (approved):
/lgtm

PR #115, opened 2024-10-01
OSD-25814 revert base image because of critical issue in the base image

tkong-redhat:
### What type of PR is this?
_(bug)_

Currently there is a critical issue in the base image which leads ocm-agent failed to start.
We need to revert the base image so ocm-agent could start properly and make the cluster back to normal status.
[OSD-25814](https://issues.redhat.com/browse/OSD-25814)
_Fixes #_


ravitri:
As discussed, it might be worth testing ocm-agent with `registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.16` image once and see if that works. If not, we can go ahead with this PR change after validation as such :+1:

PR #123, opened 2024-12-20
Bump golang.org/x/net to v0.33.0

ravitri:
### What type of PR is this?
_(bug)_

This PR updates the `golang.org/x/net` dependency to more recent version


typeid (approved):
/lgtm

PR #137, opened 2025-05-12
Calling reset metric when the SL is not been sent due to it was sent already

chamalabey:
…already

### What type of PR is this?
bug

OCMAgentResponseFailureServiceLogsSRE alert doesn't get reset for an alert which was already sent. This looks like a corner case where, service logs was sent to OCM but a network failure has caused the OCM agent to think that it failed to send the SL. 
PR changes will reset the metric when OCM-agen identifies when the SL was already sent for the firing alert. 

_Fixes #_
https://issues.redhat.com/browse/OSD-28946 

### DIFF
--- a/pkg/handlers/webhookreceiver.go
+++ b/pkg/handlers/webhookreceiver.go
@@ -125,6 +125,7 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 			log.WithFields(log.Fields{"notification": notification.Name,
 				LogFieldResendInterval: notification.ResendWait,
 			}).Info("not sending a notification as one was already sent recently")
+			metrics.ResetMetric(metrics.MetricResponseFailure)
 		} else {
 			log.WithFields(log.Fields{"notification": notification.Name}).Info("not sending a resolve notification if it was not firing or resolved body is empty")
 			s, err := managedNotifications.Status.GetNotificationRecord(notification.Name)
--- a/pkg/handlers/webhookrhobsreceiver.go
+++ b/pkg/handlers/webhookrhobsreceiver.go
@@ -192,6 +192,7 @@ func (h *WebhookRHOBSReceiverHandler) processFiringAlert(alert template.Alert, m
 		log.WithFields(log.Fields{"notification": fn.Name,
 			LogFieldResendInterval: fn.ResendWait,
 		}).Info("not sending a notification as one was already sent recently")
+		metrics.ResetMetric(metrics.MetricResponseFailure)
 		return nil
 	}
 


a7vicky:
@chamalabey Thanks for the PR. Changes look good to me. However, it's better to include alertname because multiple alerts can be covered in a single managed notification.
[example](https://github.com/openshift/managed-cluster-config/blob/master/deploy/ocm-agent-operator-managednotifications/10-managednotifications-cr.yaml)

pkg/handlers/webhookreceiver.go
```diff
@@ -168,13 +170,14 @@ func (h *WebhookReceiverHandler) processAlert(alert template.Alert, mnl *oav1alp
 		if err != nil {
 			log.WithFields(log.Fields{LogFieldNotificationName: notification.Name, LogFieldManagedNotification: managedNotifications.Name}).WithError(err).Error("unable to update notification status")
 		}
-		metrics.SetResponseMetricFailure("service_logs")
+		// Set the metric for failed service log response from OCM
+		metrics.SetResponseMetricFailure(config.ServiceLogService, notification.Name)
```
a7vicky, line 174: What do you think of including an alert name label here as well?

chamalabey: Only problem is that, the metric is not only used by service logs. It is also used when the response failures of setting Limited support flag for the clusters. Hence I have only added one label. Let me know what you think.

chamalabey: I have added an additional label as suggested.

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -168,14 +169,15 @@ func (h *WebhookRHOBSReceiverHandler) processResolvedAlert(alert template.Alert,
 			err := h.ocm.RemoveLimitedSupport(hcID, reason.ID())
 			if err != nil {
 				metrics.IncrementFailedLimitedSupportRemoved(fn.Name)
-				metrics.SetResponseMetricFailure("clusters_mgmt")
+				// Set the metric for failed limited support response from OCM
+				metrics.SetResponseMetricFailure(config.ClustersService, config.LimitedSupport)
```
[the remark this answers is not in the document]
chamalabey: I have added an additional label as suggested.

pkg/handlers/webhookrhobsreceiver.go
```diff
@@ -168,14 +169,15 @@ func (h *WebhookRHOBSReceiverHandler) processResolvedAlert(alert template.Alert,
 			err := h.ocm.RemoveLimitedSupport(hcID, reason.ID())
 			if err != nil {
 				metrics.IncrementFailedLimitedSupportRemoved(fn.Name)
-				metrics.SetResponseMetricFailure("clusters_mgmt")
+				// Set the metric for failed limited support response from OCM
+				metrics.SetResponseMetricFailure(config.ClustersService, config.LimitedSupport, alert.Labels["alertname"])
```
a7vicky, line 173: Overall looks good to me.
Metrics details suggest it should be notification name. Can we use mfn name instead `limited_support` ?

chamalabey: Done.

chamalabey:
Have tested the changes in a staging cluster to confirm the new label (_alert_name_ _notification_name_) population for the _ocm_agent_response_failure_ metric. 
Screenshot From 2025-05-28 14-01-16

ravitri:
/approve

Thanks a lot Chamal for the PR and thanks Abhishek and Tony to review!

